PR #7, opened 2019-09-20
examples/ucontext-cp: use ucontext with liburing

CarterLi:
Coroutines are often used to simplify async programming,
and they work great with liburing.
`ucontext` doesn't perform very good, but it's enough for an example.


axboe:
Thanks, this is interesting indeed, I have pulled it.

PR #10, opened 2019-09-21
Makefile generalization

rouzier:


axboe:
Thanks, that's a nice cleanup and makes it easier to add new tests.

PR #14, opened 2019-09-26
New functions

rouzier:
Added new two new function
  * io_uring_cq_ready(), returns the about of completion queue entries ready.
  * io_uring_peek_batch_cqe(), get a batch of completion queue entries.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@
 /test/b19062a56726-test
 /test/cq-full
 /test/cq-ready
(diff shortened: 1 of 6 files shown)


axboe:
Pulled, thanks. Added a commit on top to fix up some style issues in the batch patch, as well as add it to the exported list:

http://git.kernel.dk/cgit/liburing/commit/?id=6d3380256165537291377c3c0b75ed30711f22e1

PR #19, opened 2019-11-13
Issue: IORING_OP_READ_FIXED with IOSQE_IO_LINK always fails with -ECANCELED

CarterLi:
My test fails with

```
Running test fixed-link
cqe->res: Operation canceled
i = 1
fixed-link: fixed-link.c:53: main: Assertion `0' failed.
```

If I remove `io_uring_sqe_set_flags(sqe, IOSQE_IO_LINK);` at Line 39, or define `IOVECS_LEN` to 1, the test passes successfully.

```bash
$ uname -a
Linux archlinux-pc 5.3.11-arch1-1 #1 SMP PREEMPT Tue, 12 Nov 2019 22:19:48 +0000 x86_64 GNU/Linux
```


axboe:
I sent a patch to fix this, it's a small one liner. I'll get it applied for 5.4, and I'll take this test case and make it a bit more conformist to the test cases we have. Thanks for finding this!

PR #37, opened 2019-12-08
test: Fix .gitignore to omit new test binaries

bgianfo:


axboe:
Looks fine to me, but can you amend the commit to adhere to the following:

- First line should be a title. Yours already is, but it has test: in it for some reason? Looks fine if you just got rid of that.
- Commit body should have reasoning for the commit. This one is self explanatory, so you don't need that.
- Should end with a:

Signed-off-by: Real-name-of-person <email>

bgianfo:
done, thanks.

axboe:
Pulled, thanks.

PR #42, opened 2019-12-16
Fix redefinition of statx_timestamp

cor3ntin:
statx_timestamp is defined in both linux/stat.h and sys/stat.h,
causing a build failure on some systems


axboe:
Pulled, thanks.

PR #48, opened 2019-12-26
setup.c: add io_uring::features

CarterLi:
The kernel returns features it supports, but there's no way for users to get it in liburing. I add a field `io_uring::features` to store it.

It breaks ABI. Another option is make `io_uring_queue_init` return the `features` value, which still violates the document and breaks code that assumes `io_uring_queue_init` returns 0 on success.

### DIFF
--- a/src/include/liburing.h
+++ b/src/include/liburing.h
@@ -57,6 +57,7 @@ struct io_uring {
 	struct io_uring_sq sq;
 	struct io_uring_cq cq;
 	unsigned flags;
+	unsigned features;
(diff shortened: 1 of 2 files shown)


axboe:
Applications can just use io_uring_queue_init_params() to setup the ring, as that allows passing in the io_uring_params structure. Then they get immediate access to p.features afterwards. See some of the examples in test/ that test for features.

CarterLi:
I see, thanks for the explaination

PR #49, opened 2020-01-04
test: fix openat flags

wdauchy:
flags for read only should be placed in fourth position instead of fifth
one, which is for `mode`.


axboe:
Oops, thanks!

PR #61, opened 2020-01-27
Fix IORING_OP_CONNECT document

quininer:
The document for `IORING_OP_CONNECT` seems incorrect, addrlen should be stored in `off` instead of `len`.

I noticed that kernel [actually read `addr2`](https://github.com/torvalds/linux/blob/5cf9ad0e6b164a90581a59609dbf5bda3f5a089c/fs/io_uring.c#L2439), and I'm not sure which is better.


axboe:
Indeed, nice catch. Thanks!

Issue #63, opened 2020-01-31
Issue: IORING_OP_RECV returns -EFAULT constantly


CarterLi:
```c
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <unistd.h>
#include <arpa/inet.h>
#include <sys/types.h>
#include <sys/socket.h>

#include <liburing.h>

int main() {
    char buffer[1024];
    struct io_uring ring;
    struct sockaddr_in saddr;
    struct io_uring_sqe *sqe;
    struct io_uring_cqe *cqe;
    int sockfd = 0, clientfd = 0, ret;

    io_uring_queue_init(32, &ring, 0);

    sockfd = socket(AF_INET, SOCK_STREAM, 0);
    if (sockfd < 0) {
        perror("socket");
        goto err;
    }

    saddr = (struct sockaddr_in) {
        .sin_family = AF_INET,
        .sin_addr = {
            .s_addr = htonl(INADDR_ANY),
        },
        .sin_port = htons(12345),
    };

    ret = bind(sockfd, (struct sockaddr *)&saddr, sizeof(saddr));
    if (ret < 0) {
        perror("bind");
        goto err;
    }

    ret = listen(sockfd, 32);
    if (ret < 0) {
        perror("listen");
        goto err;
    }

    clientfd = accept(sockfd, NULL, NULL);
    if (clientfd < 0) {
        perror("accept");
        goto err;
    }

    sqe = io_uring_get_sqe(&ring);
    io_uring_prep_recv(sqe, clientfd, buffer, sizeof(buffer), 0);

    ret = io_uring_submit_and_wait(&ring, 1);
    if (ret <= 0) {
        perror("io_uring_submit_and_wait");
        goto err;
    }

    io_uring_peek_cqe(&ring, &cqe);

    if (cqe->res < 0) {
        printf("recv failed: %d\n", cqe->res);
        goto err;
    }

err:
    io_uring_queue_exit(&ring);
    close(clientfd);
    close(sockfd);
}
```

```shell
$ gcc recv.c -o recv -luring
$ ./recv # On another terminal: curl -v localhost:12345
recv failed: -14
$ uname -a # https://kernel.ubuntu.com/~kernel-ppa/mainline/daily/2020-01-31/
Linux carter-virtual-machine 5.5.0-999-generic #202001302109 SMP Fri Jan 31 02:15:05 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

[...]

PR #64, opened 2020-02-01
test: add test for send_recv

CarterLi:
Also ensure that `test/send_recvmsg.c` the recv thread finishs
before the process exits

Fixes https://github.com/axboe/liburing/issues/63


axboe:
Thanks!

PR #65, opened 2020-02-02
Add test for sharing uring across fork

anarazel:
Signed-off-by: Andres Freund <andres@anarazel.de>


axboe:
Thanks! This is great to have.

PR #73, opened 2020-02-12
Test suite improvements

guillemj:
This includes cleanups and fixes for issues on kernels that do not support io_uring fully,

### DIFF
--- a/test/send_recv.c
+++ b/test/send_recv.c
@@ -130,7 +130,7 @@ static int do_send(void)
 
 	ret = io_uring_queue_init(1, &ring, 0);
 	if (ret) {
-		fprintf(stderr, "queue init fail: %d\n", ret);
+		fprintf(stderr, "queue init failed: %d\n", ret);
 		return 1;
 	}
 
--- a/test/send_recv.c
+++ b/test/send_recv.c
@@ -106,13 +106,23 @@ static void *recv_fn(void *data)
 	struct io_uring ring;
 	int ret;
 
-	io_uring_queue_init(1, &ring, 0);
+	ret = io_uring_queue_init(1, &ring, 0);
+	if (ret) {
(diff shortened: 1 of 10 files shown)


test/send_recv.c
```diff
@@ -68,6 +69,10 @@ static int do_recv(struct io_uring *ring, struct iovec *iov)
 	struct io_uring_cqe *cqe;
 
 	io_uring_wait_cqe(ring, &cqe);
+	if (cqe->res == -EINVAL) {
+		fprintf(stdout, "cqe not supported, skipping\n");
```
axboe: recv not supported?

test/send_recv.c
```diff
@@ -156,6 +161,11 @@ static int do_send(void)
 	}
 
 	ret = io_uring_wait_cqe(&ring, &cqe);
+	if (cqe->res == -EINVAL) {
+		fprintf(stdout, "cqe not supported, skipping\n");
```
axboe: Ditto, but send

test/short-read.c
```diff
@@ -78,6 +78,10 @@ int main(int argc, char *argv[])
 		return 1;
 	}
 
+	if (cqe->res == -EINVAL) {
```
axboe: This one seems odd, readv has been supported since day 1?

axboe:
Just a few comments on that one patch, otherwise looks good. I'll pull it once you fix up the wording(s) in that one, thanks!

test/short-read.c
guillemj: Right sorry, this one is failing for me with cqe->res == 0, and I think I might have checked for errno == EINVAL or similar and then this mutated into that after the fact and was left over. Removed now.

axboe:
Thanks, pulled!

PR #84, opened 2020-02-22
Added and renamed parameters to `flags`

YoSTEALTH:
Added `flags` parameter to  `io_uring_prep_readv` and `io_uring_prep_writev`. Also changed `fsync_flags` to `flags` in `io_uring_prep_fsync` to stay consistent with other prep function parameters.


axboe:
It’s unfortunate that it doesn’t follow the later adopted convention of passing on flags, but let’s not make an incompatible change just for this. It’ll break compilation of existing apps. Does it even build liburing with this change?

YoSTEALTH:
Sorry, I am not really a C programmer, I thought I was helping out by adding what might be a missing feature. Figured there would be build-bot on here. Can delete this? Thanks.

axboe:
No worries, no harm in trying! The build bot is a bit broken, haven't had time to look into that yet.

PR #90, opened 2020-03-05
io_uring_queue_init will hang if the `entries=0`

YoSTEALTH:
since `io_uring_queue_init` calls `io_uring_queue_init_params` internally, added code in `io_uring_queue_init_params` only.

### DIFF
--- a/src/setup.c
+++ b/src/setup.c
@@ -134,6 +134,9 @@ int io_uring_queue_init_params(unsigned entries, struct io_uring *ring,
 			       struct io_uring_params *p)
 {
 	int fd, ret;
+	
+	if (entries < 1)
+		return -1;
 
 	fd = __sys_io_uring_setup(entries, p);
 	if (fd < 0)


axboe:
Hmm not sure I follow, it'll return -EINVAL if you ask for 0 entries. What do you mean "it hangs"?

YoSTEALTH:
O i c, I been getting `Segmentation fault (core dumped)` i will investigate

axboe:
That's probably because you're ignoring the return value, and then proceed to use the ring? The contents of (most of) the ring would be uninitialized if the setup system call fails.

axboe:
Just for completeness, the patch wasn't right either. You must return -errno, which meant your patch would return -EPERM. But the system call catches this and returns -1/-EINVAL, which io_uring_queue_init() then passes back.

YoSTEALTH:
try:
    no = io_uring_queue_init(0, ring, 0)
    if no < 0:
        raise OSError(-no, os.strerror(-no))
finally:
    io_uring_queue_exit(ring)


Woo never seen anything do this before.

I would catch the return value and raise python exception according to what the return value was. For example it would raise `BlockingIOError` if return value was `-11`. Finally it would call `io_uring_queue_exit(ring)` to close the `ring`. So if the `entries=0` the python exception would somehow get suppressed by calling `io_uring_queue_exit(ring)` thus core-dump would happen.

Normal behavior for python is to raise another exception.

Thanks for your comments.

axboe:
Ah I see, thanks for the follow up!

PR #94, opened 2020-03-08
Use __atomic builtins in barrier.h for better architecture support

taylordotfish:
`io_uring_smp_store_release()` and `io_uring_smp_load_acquire()` currently issue full sequentially consistent memory barriers (as opposed to release and acquire barriers, respectively) on every architecture except x86(-64) and 64-bit ARM.

These macros can instead be implemented using [GCC's `__atomic` builtins](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html), which were [introduced in GCC 4.7](https://gcc.gnu.org/gcc-4.7/changes.html). This ensures that the proper, most optimal instructions are used for every architecture that GCC supports.

Given that io_uring first appeared in Linux 5.1, it seems very unlikely that anyone will need to compile liburing with any version of GCC prior to 4.7, which was released in 2012.

I haven't observed any regressions on my machine, although my kernel doesn't have full support for io_uring.

### DIFF
--- a/src/include/liburing/barrier.h
+++ b/src/include/liburing/barrier.h
@@ -34,19 +34,6 @@ after the acquire operation executes. This is implemented using
 
 #if defined(__x86_64__) || defined(__i386__)
 /* Adapted from arch/x86/include/asm/barrier.h */
-#define io_uring_mb()		asm volatile("mfence" ::: "memory")
-#define io_uring_rmb()		asm volatile("lfence" ::: "memory")
-#define io_uring_wmb()		asm volatile("sfence" ::: "memory")
-#define io_uring_smp_rmb()	io_uring_barrier()
-#define io_uring_smp_wmb()	io_uring_barrier()
-#if defined(__i386__)
-#define io_uring_smp_mb()	asm volatile("lock; addl $0,0(%%esp)" \
-					     ::: "memory", "cc")
-#else
-#define io_uring_smp_mb()	asm volatile("lock; addl $0,-132(%%rsp)" \
-					     ::: "memory", "cc")
-#endif
-
 #define io_uring_smp_store_release(p, v)	\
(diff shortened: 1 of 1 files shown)


axboe:
Let's please do that for the ones where we don't currently have them defined, I don't want to do it wholesale.

taylordotfish:
> Let's please do that for the ones where we don't currently have them defined, I don't want to do it wholesale.

Okay, I've made this change. Out of curiosity, what is the rationale for handling certain architectures explicitly?

axboe:
I have higher faith in what we have than being at the mercy of a (potentially) old compiler. Lower risk change just to do it for archs we don't currently have actual "support" for.

axboe:
Do feel free to kill the unused mb/rmb/wmb stuff though.

taylordotfish:
I'm not convinced that the AArch64 implementation is optimal. We're currently issuing a `dmb ish` for both loads and stores, but `dmb ish` is essentially a full barrier: it prevents loads and stores from being reordered across the barrier in both directions, which is overkill when we just need a load with acquire semantics or a store with release semantics.

I think `stlr` and `ldar` are better instructions for stores and loads, respectively, and this is exactly what [arch/arm64/include/asm/barrier.h](https://github.com/torvalds/linux/blob/378fee2e6b12f31ab3749e0aa4ed0a63be23e822/arch/arm64/include/asm/barrier.h#L76) does,  which is supposedly where the AArch64 implementation was copied from.

axboe:
I'm fine with going with builtins for aarch64 as well.

taylordotfish:
I rebased onto master; merging this PR will now remove the unused mb/rmb/wmb code.

axboe:
Looks good to me! Pulled, thanks.

PR #104, opened 2020-04-10
Clarify licensing

guillemj:
This series try to clarify the current licensing situation, with the objective of then being able to upload this to Debian. I've tried to infer the correct licenses from the stuff in the repo as a starting point, but probably my guesses are wrong, I'm happy to amend any of them.

I think the spec file syntax is correct, but I've not run rpmlint.

I've also left the debian directory alone for now which contains incorrect licensing data, with invalid syntax, as packaged this project from scratch anyway, and I'm not sure whether there's value in shipping the packaging upstream? So perhaps a better option is to remove it? But if you want to preserve it, I could prepare a targeted fix for just the copyright file though.

### DIFF
--- a/COPYING.GPL
+++ b/COPYING.GPL
@@ -1,0 +1,339 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 2, June 1991
+
+ Copyright (C) 1989, 1991 Free Software Foundation, Inc.,
+ 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
+  The licenses for most software are designed to take away your
+freedom to share and change it.  By contrast, the GNU General Public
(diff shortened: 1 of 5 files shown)


src/include/liburing/io_uring.h
```diff
@@ -1,4 +1,4 @@
-/* SPDX-License-Identifier: (GPL-2.0 WITH Linux-syscall-note) OR MIT */
+/* SPDX-License-Identifier: (GPL-2.0-only WITH Linux-syscall-note) OR MIT */
```
axboe, line 1: I'd rather not make this change unless I absolutely have to, since the header is synced with the kernel and then I have to make that change there too.

axboe:
Can you drop the io_uring.h change, then I'm happy with the rest.

src/include/liburing/io_uring.h
guillemj: Ah ok, makes sense, I guess this would need to be fixed in the kernel then, but that might imply lots of instances being touched. I'll pull this one out then.

axboe:
Thanks, pulled.

PR #113, opened 2020-04-23
test: Use __NR_mmap2 when __NR_mmap is not defined

guillemj:
This was found on the Debian build daemons <https://buildd.debian.org/status/logs.php?pkg=liburing&ver=0.6-1> once the package got accepted in the archive. Revision 0.6-2 includes this patch and builds now everywhere where it has been built <https://buildd.debian.org/status/package.php?p=liburing>. :)

On some new 32-bit architectures (such as armel and armhf), __NR_mmap
is not defined, but we can safely use __NR_mmap2 as we are passing an
offset of 0, so the different bases for that argument do not apply.

We need to stop defining __NR_mmap in one of the unit tests to be able
to properly fallback to the __NR_mmap2 case.

### DIFF
--- a/test/35fa71a030ca-test.c
+++ b/test/35fa71a030ca-test.c
@@ -260,8 +260,14 @@ static void loop(void)
 #ifndef __NR_io_uring_setup
 #define __NR_io_uring_setup 425
 #endif
-#ifndef __NR_mmap
-#define __NR_mmap 192
+
+/* We can use the same syscall, because our offset is 0. */
+#if defined(__NR_mmap)
+#define SYSCALL_mmap __NR_mmap
+#elif defined(__NR_mmap2)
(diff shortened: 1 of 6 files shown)


axboe:
This is all for generated code from syzbot. Why not just make it use mmap() instead of doing a direct syscall?

axboe:
Does this fix it:

https://git.kernel.dk/cgit/liburing/commit/?id=d7720beb19790da00f44c7583340d29d60520490

guillemj:
Right that is certainly better. I thought about it, but was not sure whether the intention with these tests was to have them being as minimal as possible. Should have asked I guess. :) Thanks, I'm assuming this will also fix the issue, so will pull that in on the next revision upload.

PR #120, opened 2020-04-29
Adding description for the IORING_OP_SPLICE operation

shuveb:
Add details of the splice operation to the man page.


axboe:
Thanks

PR #130, opened 2020-05-07
test/lfs-openat: open dfd with O_PATH

MaxKellermann:
This triggers a bug present until at least Linux 5.6.11, causing
EBADF, see https://lkml.org/lkml/2020/5/7/1287

Signed-off-by: Max Kellermann <mk@cm4all.com>


MaxKellermann:
Rebased after @axboe's force-push.

PR #149, opened 2020-07-02
man: fix description of splice offsets

gerow:
The interface to splice(2) in io_uring differs in that it doesn't accept a
pointer for the offset values, but instead a literal loff_t. It then uses a
sentinel value of -1 to pass the equivalent of a NULL. This fixes
io_uring_enter.2 to clarify this.


axboe:
Thanks!

PR #152, opened 2020-07-04
Take `sockaddr` immutably in io_uring_prep_connect

### LINKED ISSUE #151: `const` correctness in `io_uring_prep_connect`
At least one of the APIs differs in its handling of `const`, that is `io_uring_prep_connect`. The underlying API takes the `sockaddr` as `const`, but the io_uring API takes it mutably.

* io_uring API: https://github.com/axboe/liburing/blob/master/src/include/liburing.h#L325
* connect API: https://man7.org/linux/man-pages/man2/connect.2.html

I could do a PR for this if it's something that can be fixed. I guess the documentation web site also needs to be updated (where would that fix need to be directed?)

KayEss:
This is the same as the underlying `connect` system call and fixes #151

### DIFF
--- a/src/include/liburing.h
+++ b/src/include/liburing.h
@@ -322,7 +322,7 @@ static inline void io_uring_prep_link_timeout(struct io_uring_sqe *sqe,
 }
 
 static inline void io_uring_prep_connect(struct io_uring_sqe *sqe, int fd,
-					 struct sockaddr *addr,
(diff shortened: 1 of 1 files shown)


axboe:
You forgot the man page, but I did that separately.

KayEss:
Sorry, that was not my intention. I didn't find the function name in the man pages. I guess I just don't understand the file format.

PR #156, opened 2020-07-07
compat_libaio.c: get correct io_event data from cqe->user_data

nan1994:
cqe->user_data is a point to struct io_completion,not struct iocb.

Signed-off-by: Hongnan Li <hongnan.li@linux.alibaba.com>

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -24,7 +24,7 @@ include ../Makefile.quiet
 
 all: $(all_targets)
 
-liburing_srcs := setup.c queue.c syscall.c register.c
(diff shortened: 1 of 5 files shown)


axboe:
I'm not really maintaining the wrapper, it was more of a thought experiment to see if it could work. Are you actively using it? Main reason I dropped it is that we'd need some kernel support to enable the user mapped ring to work, and I didn't want to pile more stuff on there. But if you're just using the syscalls for IO with aio, then it can be made complete.

axboe:
I applied the patch and rebased on top of current master, just in case.

PR #162, opened 2020-07-16
man/io_uring_enter: fix openat sqe doc

necipfazil:


YoSTEALTH:
@necipfazil  this change has already been done, maybe you were looking at older branch.

axboe:
@YoSTEALTH no it hasn't

YoSTEALTH:
O, my bad, I was looking at the wrong reference :)

PR #167, opened 2020-07-22
fix debian packaging

metze-samba:


axboe:
That was quick, thanks Stefan!

PR #178, opened 2020-08-23
Write that follows timed out write always hangs

dshulyak:
Write on the same fd that follows write that was canceled with IORING_OP_LINK_TIMEOUT never completes. 
See attached tests for repro.

Was discovered on 5.7.15

### DIFF
--- a/test/link-timeout.c
+++ b/test/link-timeout.c
@@ -992,6 +992,106 @@ err:
 	return 1;
 }
 
+static int test_write_after_timeout_hangs(struct io_uring *ring)
+{
+	struct __kernel_timespec ts;
+	struct io_uring_cqe *cqe;
+	struct io_uring_sqe *sqe;
+	int fd, ret, i;
+	char *file;
+	struct iovec iov;
+	char buffer[128];
+
+	file = ".test_timeout_hangs";
+	fd = open(file, O_WRONLY | O_CREAT, 0644);
+	if (fd < 0) {
+		perror("open file");
+		goto err;
+	}
+	sqe = io_uring_get_sqe(ring);
+	if (!sqe) {
+		printf("get sqe failed\n");
+		goto err;
+	}
+
+	iov.iov_base = buffer;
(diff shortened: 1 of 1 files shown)


isilence:
This one was pretty useful. Are you going to reopen it later?
I'll comment on a coupe of thing there anyway.

test/link-timeout.c
```diff
@@ -992,6 +992,106 @@ static int test_timeout_link_chain5(struct io_uring *ring)
...
+			if (cqe->res != -ETIME) {
+				fprintf(stderr, "got %d, wanted -ETIME\n", cqe->res);
+				goto err;
+			}
+			break;
+		}
+		io_uring_cqe_seen(ring, cqe);
+	}
+
+	fprintf(stderr, "timeout + write were completed\n");
```
isilence, line 1064: Please, remove non-error printf()'s

test/link-timeout.c
```diff
@@ -992,6 +992,106 @@ static int test_timeout_link_chain5(struct io_uring *ring)
...
+
+		switch (cqe->user_data) {
+		case 1:
+			if (cqe->res != -ECANCELED) {
+				fprintf(stderr, "write got %d, wanted -ECANCELED\n", cqe->res);
+				goto err;
+			}
+			break;
+		case 2:
+			if (cqe->res != -ETIME) {
```
isilence, line 1055: The write may actually complete before the timeout had a chance
to cancel it, so it fails for me from time to time.

dshulyak: i noticed this as well. i have tried to use timespec with s=0, ns=0, but it doesn't seem to work as i expected

axboe: -EALREADY is probably also a valid return here.

axboe:
> This one was pretty useful. Are you going to reopen it later?

Agree, I think we should add the test case, just needs a bit of cleaning to adhere to what is expected of the regression tests.

PR #193, opened 2020-09-06
Fix io_uring_sq_ready() to use load acquire to read khead

jorangreef:
Without the memory barrier, we could miss an update and think the SQ
wasn't ready when in fact it was.

### DIFF
--- a/src/include/liburing.h
+++ b/src/include/liburing.h
@@ -458,6 +458,13 @@ static inline void io_uring_prep_shutdown(struct io_uring_sqe *sqe, int fd,
  */
 static inline unsigned io_uring_sq_ready(struct io_uring *ring)
 {
+	/*
+	 * Without a barrier, we could miss an update and think the SQ wasn't ready.
+	 * We don't need the load acquire for non-SQPOLL since then we drive updates.
+	 */
+	if (ring->flags & IORING_SETUP_SQPOLL)
+		return ring->sq.sqe_tail - io_uring_smp_load_acquire(ring->sq.khead);
+
 	/* always use real head, to avoid losing sync for short submit */
 	return ring->sq.sqe_tail - *ring->sq.khead;
 }


src/include/liburing.h
```diff
@@ -459,7 +459,7 @@ static inline void io_uring_prep_shutdown(struct io_uring_sqe *sqe, int fd,
 static inline unsigned io_uring_sq_ready(struct io_uring *ring)
 {
 	/* always use real head, to avoid losing sync for short submit */
-	return ring->sq.sqe_tail - *ring->sq.khead;
+	return ring->sq.sqe_tail - io_uring_smp_load_acquire(ring->sq.khead);
```
axboe, line 462: Since this is only strictly needed for SQPOLL, maybe we should make it dependent on that? Something ala:

```
if (!(ring->flags & IORING_SETUP_SQPOLL))
    return ring->sq.sqe_tail - io_uring_smp_load_acquire(ring->sq.khead);
return ring->sq.sqe_tail - *ring->sq.khead;
```

Totally untested, just typed it in here... Then probably also include a comment as to why.

BTW, please include a Signed-off-by: Real Name <email> at the end of your commit. Thanks!

jorangreef: Thanks, done. The bitwise condition was fine but needed to be flipped around.

axboe:
Thanks! Please fold the whitespace commit with the one that introduces it, no point having that muddy up the history. Also please include your email in the signed-off-by, it should look like the ones in the git log:

jorangreef:
The commits have been squashed (this could also be done by GitHub when merging the PR?) and my email address has been added.

axboe:
> this could also be done by GitHub when merging the PR?

I don't merge via GH, and GH isn't the canonical repo for liburing. It just pushes it here as a backup.

PR #194, opened 2020-09-07
man/io_uring_setup.2: add rationale for smp_mb() before wakeup

jorangreef:
Signed-off-by: Joran Dirk Greef <joran@coil.com>


axboe:
Thanks for sending this in. I think we should change this to indicate load acquire semantics. A full memory barrier is not really needed here. Let me know if you just want me to make that documentation change, or if you want to update this one.

jorangreef:
Sounds good. All yours! 👍

PR #196, opened 2020-09-09
probes: provide our own free function

glommer:
liburing has a convenience function to get a io_uring_probe that can
be used to test for valid features. The function returns a probe that
the caller has to then deallocate.

That's not a problem if we are dealing with languanges like C and C++
but when trying to uring liburing bindings in languages with a more
well defined barrier into the C interface, calling free() can be at
best cumbersome and at worst just wrong as the language may be using
a different allocator under the hood.

Since we provide an allocated version of probe, provide users with
a way to deallocate it.

### DIFF
--- a/src/include/liburing.h
+++ b/src/include/liburing.h
@@ -77,6 +77,12 @@ extern struct io_uring_probe *io_uring_get_probe_ring(struct io_uring *ring);
 /* same as io_uring_get_probe_ring, but takes care of ring init and teardown */
 extern struct io_uring_probe *io_uring_get_probe(void);
 
+/*
+ * frees a probe allocated through io_uring_get_probe() or
+ * io_uring_get_probe_ring()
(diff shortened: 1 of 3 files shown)


axboe:
Should go into the symbol map as well.

glommer:
ah, you are right. Sorry about that. How are you handling versioning of the patches here, Jens ? Should I just force-push into the same branch ?

axboe:
Yeah just force push, that's totally fine.

glommer:
done

axboe:
Thanks! Just now noticed that you're missing a Signed-off-by in the commit message. Can you add that, too?

glommer:
ok, I can't keep track anymore of which projects require SoB and which don't =)

Added

axboe:
Thanks!

PR #200, opened 2020-09-14
Adding section 7 man page for io_uring

shuveb:
Draft version of the section 7 man page for io_uring.


axboe:
This is great! Just a heads up that I've pulled it in, I'm making some edits and I'll push it out tomorrow. Figured that was the best way to keep this rolling.

shuveb:
Thanks, Jens!

Issue #190, opened 2020-09-03
Feature Request: API Reference Documentation


pcewing:
User story:
> As a user of `liburing` I would like API reference documentation in the form of
> man-pages so that the library is easier to use correctly.

I'm happy to help write these as I gain familiarity with the library.

### Progress

Not sure if we really need a man page for all of these but I just grabbed the list of functions that are externed in `liburing.h`.

- [ ] io_uring_get_probe_ring
- [ ] io_uring_get_probe
- [ ] io_uring_free_probe
- [ ] io_uring_queue_init_params
- [x] io_uring_queue_init
- [ ] io_uring_queue_mmap
- [ ] io_uring_ring_dontfork
- [x] io_uring_queue_exit
- [ ] io_uring_wait_cqes
- [ ] io_uring_wait_cqe_timeout
- [ ] io_uring_submit
- [ ] io_uring_submit_and_wait
- [x] io_uring_get_sqe
- [ ] io_uring_register_buffers
- [ ] io_uring_unregister_buffers
- [ ] io_uring_register_files
- [ ] io_uring_unregister_files
- [ ] io_uring_register_files_update
- [ ] io_uring_register_eventfd
- [ ] io_uring_register_eventfd_async
- [ ] io_uring_unregister_eventfd
- [ ] io_uring_register_probe
- [ ] io_uring_register_personality
- [ ] io_uring_unregister_personality
- [ ] io_uring_register_restrictions
- [ ] io_uring_enable_rings
- [ ] __io_uring_sqring_wait
- [ ] __io_uring_get_cqe

axboe:
Not sure why this was closed? The system calls are pretty decently documented in the man pages included with liburing, but not the liburing functions themselves. I know an io_uring(7) man page is being worked on, but would love to see something for liburing.

pcewing:
I closed this because it looked like some man pages were already written; however, you're correct that they only cover the io_uring system calls and not `liburing`. I'll reopen this with the more focused feature request of liburing api reference man-pages.

sesse:
A related request: It would be very useful to have comments for each function in the .h file. I see that there are already many comments on the implementations that could be used for this purpose (copied or moved); e.g. src/queue.c has comments for functions like io_uring_get_sqe(), but a typical developer will be looking in the .h file (not the least because that's the only thing available in most -dev/-devel packages). Optionally, some of the excellent documentation from LoTiouR could probably also find a good home here. :-)

axboe:
[...]

PR #202, opened 2020-09-16
Add section 3 man pages for init and exit

pcewing:
This PR is to add the first two API reference man pages as requested in
https://github.com/axboe/liburing/issues/190

It may take a while to get through all of these so I think it makes more sense
to add them piecemeal.

Note the `TODO` comments. I looked at how another library (libcurl) formats its
title section and they put the release version and release date; however, that
would be a pain to keep updated without proper tooling. Let me know how you
would prefer that to be formatted for liburing.

Also, I'm not familiar with Debian packaging but did my best to update that as
well; it's currently untested as I'll need to figure out how to do that.


axboe:
Sorry for being slow on this one - great started, applied!

PR #206, opened 2020-09-20
Build system fixes

guillemj:
Fixes build flags settings, by honoring CPPFLAGS, and matches the CXXFLAGS handling with the other flags. In addition adds a missing ignore into .gitignore.

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -1,6 +1,8 @@
+CPPFLAGS ?=
+override CPPFLAGS += -D_GNU_SOURCE -I../src/include/
 CFLAGS ?= -g -O2
 XCFLAGS =
-override CFLAGS += -Wall -D_GNU_SOURCE -L../src/ -I../src/include/
+override CFLAGS += -Wall -L../src/
 
 include ../Makefile.quiet
 
@@ -17,7 +19,7 @@ test_srcs := io_uring-test.c io_uring-cp.c link-cp.c
 test_objs := $(patsubst %.c,%.ol,$(test_srcs))
 
 %: %.c
-	$(QUIET_CC)$(CC) $(CFLAGS) -o $@ $< -luring $(XCFLAGS)
+	$(QUIET_CC)$(CC) $(CPPFLAGS) $(CFLAGS) -o $@ $< -luring $(XCFLAGS)
 
 clean:
 	@rm -f $(all_targets) $(test_objs)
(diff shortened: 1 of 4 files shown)


examples/Makefile
```diff
@@ -1,6 +1,8 @@
+CPPFLAGS ?=
```
isilence, line 1: Where exactly are CPPFLAGS used in liburing?
AFAIR It doesn't depend on anything to propagate them.

examples/Makefile
```diff
@@ -17,7 +19,7 @@ test_srcs := io_uring-test.c io_uring-cp.c link-cp.c
 test_objs := $(patsubst %.c,%.ol,$(test_srcs))
 
 %: %.c
-	$(QUIET_CC)$(CC) $(CFLAGS) -o $@ $< -luring $(XCFLAGS)
+	$(QUIET_CC)$(CC) $(CPPFLAGS) $(CFLAGS) -o $@ $< -luring $(XCFLAGS)
```
isilence, line 22: I don't find it a good practice passing to CC anything CPP related (i.e. CPPFLAGS).
Better to decouple them

guillemj: Well CC in this case is performing the tasks of the preprocessor, the compiler and the linker, so I don't see why it would not make sense to honor the flags. Notice the options moved to the CPPFLAGS are really preprocessor related options.
In the end CC here is acting as a frontend driving the various phases.

Also notice the default make rules which do honor these various flags:
```shell
make -p -f /dev/null | grep CPPFLAGS
```

Just realized too that for linking invocations we are not passing LDFLAGS,

examples/Makefile
guillemj: They are not currently used, that's the point of this patch. :)

examples/Makefile
guillemj: Or from another point of view. I think it is a common expectation that a user can set these flags when building a project and they will affect any invocation involving their build phase. So if as a user I specify say any of CPPFLAGS or CFLAGS or LDFLAGS, then any invocation that would end up calling the relevant backends should end up being affected by them, regardless of these being called as CC, CPP, LD, etc.

isilence: Ohh blood, I was looking and CXXFLAGS and have got confused in the end.
CPPFLAGS there look sensible, as well as does the whole patch.
LGTM, thanks!

PR #230, opened 2020-10-27
Moves function calls out of assert() to run tests outside debug mode.

sharpelletti:
assert() is redefined as check() in liburing.h to ensure that it
will run if NDEBUG is defined. Without this, certain tests will
fail due to functions being called within an assert(), which are
skipped if NDEBUG is defined.
    
check() is a macro that will use assert() if NDEBUG is not defined.
Otherwise, it uses its own implementation.

### DIFF
--- a/test/232c93d07b74-test.c
+++ b/test/232c93d07b74-test.c
@@ -60,20 +60,25 @@ static void *rcv(void *arg)
 {
 	struct params *p = arg;
 	int s0;
+	int res;
 
 	if (p->tcp) {
 		int val = 1;
+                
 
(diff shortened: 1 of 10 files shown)


axboe:
Yeah good point, it's all from contributed cases, I generally dislike debug/assert statements with side effects. While this would work, I'd greatly prefer having them all changed to something ala:

ret = foo();
assert(foo != 0);

or whatever is appropriate for the given case. Any chance I can talk you into doing that change instead?

axboe:
Thanks! Can you squash the two commits into one?

axboe:
Thanks!

PR #238, opened 2020-11-17
Changes format string from %lld to %PRIu64 in several tests.

sharpelletti:
6 liburing tests failed to build on powerpc due to a 'long' versus
'long long' mismatch. Changing the identifer from %lld to %PRIu64
and casting the value to uint64_t fixes this.


axboe:
Pulled, thanks.

PR #251, opened 2020-12-09
man page missing installation

maze888:
Hi.

I'm a person who is interested in liburing

I added manual page installation


axboe:
Can you turn that into one clean commit?

PR #260, opened 2020-12-14
a fix for an io_uring_setup manual

wapxmas:


axboe:
Thanks for the fix! Two changes I'd like to see:

- You need to write a proper commit message, and also include your signed-off-by at the end. Look at other commits in the repo for what the format should look like.

- You have a merge in there, get rid of that. The pull request should just be your change(s), not an unrelated merge.

PR #264, opened 2020-12-16
Add __builtin_expect(cond, 0) to NEED_WAKEUP for SQPOLL.

goldsteinn:
If heavy IO load then kernel thread will only very rarely need to be woken up. Only case where this condition is common is if timeout expected between IO calls in which case should be using GETEVENTS.

Have seen this condition get optimized for (compiler puts it as the fall through case).

Possible worth considering wrapping some of the error checking with this?

### DIFF
--- a/src/include/liburing.h
+++ b/src/include/liburing.h
@@ -15,6 +15,14 @@
 #include "liburing/io_uring.h"
 #include "liburing/barrier.h"
 
+#ifndef uring_unlikely
+#  define uring_unlikely(cond)      __builtin_expect(!!(cond), 0)
+#endif
+
+#ifndef uring_likely
+#  define uring_likely(cond)        __builtin_expect(!!(cond), 1)
+#endif
+
 #ifdef __cplusplus
 extern "C" {
 #endif
--- a/src/queue.c
+++ b/src/queue.c
@@ -25,10 +25,11 @@ static inline bool sq_ring_needs_enter(struct io_uring *ring, unsigned *flags)
 {
 	if (!(ring->flags & IORING_SETUP_SQPOLL))
 		return true;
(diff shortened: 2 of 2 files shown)


axboe:
Change looks fine to me, but your commit message does not. What you put in the description here should be the commit message! You also need to add your Signed-off-by: at the end.

liburing commit messages are generally:

- First line the title
- Empty line
- Then commit message
- Empty line
- Signed-off-by: Author <email>

Here's an example:

```
commit 516280a50fbcc4e6330ec20f6c93128caff510c4
Author: Pavel Begunkov <asml.silence@gmail.com>
Date:   Tue Dec 8 21:57:29 2020 +0000

    rem_buf/test: inital testing for OP_REMOVE_BUFFERS
    
    Basic testing for IORING_OP_REMOVE_BUFFERS. test_rem_buf(IOSQE_ASYNC)
    should check that it's doing locking right when punted async.
    
    Signed-off-by: Pavel Begunkov <asml.silence@gmail.com>
```

Can you fix up the commit and let me know? Then I can pull it.

goldsteinn:
Updated the commit message. What are your thoughts on wrapping some of the error checking that could be on the critical path with unlikely?

axboe:
As long as it's done sanely, then I think it's a good idea. This patch is a good example, since the fast path is indeed not needing a system call. So even if we mispredicted and it turns out all calls do need to enter the kernel, we haven't really lost anything as that's the vastly more expensive path.

I just don't want to pepper likely/unlikely all over the place, only where they make sense and potentially make a difference.

axboe:
I pulled this one, but one more thing for commit messages: please wrap them at 72-74 chars, makes them a lot more readable when doing git log. Just keep it in mind for future changes, thanks.

PR #282, opened 2021-01-23
Fix IORING_OP_FALLOCATE args in io_uring_enter(2) man page

jorangreef:
`addr` must in fact contain the length.
`len` is used for the fallocate mode.

See: https://patchwork.kernel.org/project/linux-fsdevel/patch/20191213183632.19441-2-axboe@kernel.dk/

Signed-off-by: Joran Dirk Greef <joran@coil.com>


axboe:
Indeed, good catch. I'll pull this, thanks.

PR #284, opened 2021-01-26
Adding support for skipping files in kernel patch. This will add tests / documentation for said patch.

goldsteinn:
This is to add testing / documentation for patch: ```0001-io_uring-Add-skip-option-for-__io_sqe_files_update.patch``` if it is applied.

Should not be accepted unless the patch is as well.

### DIFF
--- a/man/io_uring_register.2
+++ b/man/io_uring_register.2
@@ -144,6 +144,7 @@ This operation replaces existing files in the registered file set with new
 ones, either turning a sparse entry (one where fd is equal to -1) into a
 real one, removing an existing entry (new one is set to -1), or replacing
 an existing entry with a new existing entry.
+
 .I arg
 must contain a pointer to a struct io_uring_files_update, which contains
 an offset on which to start the update, and an array of file descriptors to
@@ -151,6 +152,12 @@ use for the update.
 .I nr_args
 must contain the number of descriptors in the passed in array. Available
 since 5.5.
+
+File descriptors can be skipped if they are set to
+.B IORING_REGISTER_FILES_SKIP.
+Skipping an fd will not touch the file assosiated with the previous
+fd at that index. Available since 5.12.
+
 
 .TP
 .B IORING_UNREGISTER_FILES
--- a/test/file-register.c
+++ b/test/file-register.c
@@ -550,6 +550,44 @@ err:
 	return 1;
 }
 
+static int test_skip(struct io_uring *ring)
+{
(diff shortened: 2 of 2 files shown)


isilence:
The mystery for me is where the kernel patch itself...

axboe:
It was posted. Seems like lore/vger has some issues these days, so slow...

goldsteinn:
> The mystery for me is where the kernel patch itself... 

its not on lore yet but should be visible through email ccd to:
```
It should be most recently patch 
axboe@kernel.dk,
viro@zeniv.linux.org.uk,
io-uring@vger.kernel.org,
linux-fsdevel@vger.kernel.org,
linux-kernel@vger.kernel.org,
goldstein.w.n@gmail.com
```

I'll add you directly as a cc in the future.

For now forwarded it to you.

isilence:
*Delivered after 9630 seconds*, nice...

Looks like a good occasion to finally add myself into /MAINTEINERS.
How about removing io_uring from under Al/fsdevel?

axboe:
I'll add you to maintainers. But yes, it's really annoying why it keeps adding Al when you just use the maintainers script. Not sure there's a way to solve that without moving it into kernel/io_uring/ (which would be more appropriate than fs/io_uring/).

axboe:
> I'll add you to maintainers

Actually just send a patch for that yourself :-)

isilence:
`X:` exclude from FILESYSTEMS? It works, but kind of...

axboe:
Yeah, it'd work for that entry. Arguably io_uring should just be moved, but let's just work around it instead.

isilence:
> I'll add you to maintainers. But yes, it's really annoying why it keeps adding Al when you just use the maintainers script. Not sure there's a way to solve that without moving it into kernel/io_uring/ (which would be more appropriate than fs/io_uring/).

Since I'm not committing directly, reviewer (aka R:) makes more sense, don't you think?

axboe:
Sure, that works for me. It means you'll get CC'ed on patches, which would eliminate annoying delays like today.

isilence:
> Yeah, it'd work for that entry. Arguably io_uring should just be moved, but let's just work around it instead.

Agree, bad for history and not doesn't matter anyway.

axboe:
git would track the history fine, but it'd be a pain for backports etc. No point in having such a flag day, location doesn't really matter if we can fix this maintainers annoyance.

PR #302, opened 2021-02-19
setup: check whether malloc succ before using it

ZhiqiangLiu26:
In io_uring_get_probe_ring func, we will call malloc()
to alloc memory for probe. Considering malloc may return NULL,
we should check whether malloc() succ before using it.


axboe:
Pulled, thanks.

PR #303, opened 2021-02-20
tests: call assert() when allocating memory fails

ZhiqiangLiu26:
In tests, we can just call assert() when allocating memory fails,
as done in io_uring_regeister.c and io_uring_enter.c.

### DIFF
--- a/test/accept.c
+++ b/test/accept.c
@@ -7,7 +7,6 @@
 #include <stdlib.h>
 #include <stdint.h>
 #include <assert.h>
-
 #include <errno.h>
 #include <fcntl.h>
 #include <unistd.h>
@@ -33,6 +32,7 @@ static void queue_send(struct io_uring *ring, int fd)
 	struct data *d;
 
 	d = malloc(sizeof(*d));
+	assert(d);
 	d->iov.iov_base = d->buf;
 	d->iov.iov_len = sizeof(d->buf);
 
@@ -46,6 +46,7 @@ static void queue_recv(struct io_uring *ring, int fd)
 	struct data *d;
 
 	d = malloc(sizeof(*d));
+	assert(d);
 	d->iov.iov_base = d->buf;
 	d->iov.iov_len = sizeof(d->buf);
 
--- a/test/cq-overflow.c
+++ b/test/cq-overflow.c
@@ -9,6 +9,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <fcntl.h>
+#include <assert.h>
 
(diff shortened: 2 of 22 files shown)


axboe:
It makes more sense to do this in the lib itself, for tests we can just let them crash. The kernel usually grinds itself into the ground before it fails to hand out memory.

Is this a setup concern where you regularly see malloc failures and expect the app to handle it? Because if not, I'd much rather just leave it be rather than have this churn. Crashing on NULL pointer deref or explicitly with an assert, doesn't make a lot of difference for the test case.

axboe:
Another concern is that new tests will forget. Here's what I think _could_ be done if you wanted to harden the tests:

- We should have some helper code that can be used from tests. Right now a lot of basic stuff is duplicated, and it makes the tests bigger than they should be. If we had a helpers.c and helpers.h that had various utilities that tests could use, then that'd help make tests simpler.
- If we had that helper, then you could add memory allocation helpers and convert the tests to use them.
- If you had memory alloc helpers, then you have one unified spot to put the NULL checking into. The callers would not need to check, as they'd either get memory or an assert would be triggered and the call never returns.

If you were willing to start this helpers.c part with adding the mem alloc routines we use, then I'd be all for doing the assert. Putting the logic and assert in each caller, not so much.

ZhiqiangLiu26:
> Another concern is that new tests will forget. Here's what I think _could_ be done if you wanted to harden the tests:
> 
> * We should have some helper code that can be used from tests. Right now a lot of basic stuff is duplicated, and it makes the tests bigger than they should be. If we had a helpers.c and helpers.h that had various utilities that tests could use, then that'd help make tests simpler.
> * If we had that helper, then you could add memory allocation helpers and convert the tests to use them.
> * If you had memory alloc helpers, then you have one unified spot to put the NULL checking into. The callers would not need to check, as they'd either get memory or an assert would be triggered and the call never returns.
> 
> If you were willing to start this helpers.c part with adding the mem alloc routines we use, then I'd be all for doing the assert. Putting the logic and assert in each caller, not so much.

thanks for your patient reply. I will try to write this helper as you said.
Actually, I also found there were many duplicated codes in tests. After adding helper, I will try to do some cleanup.

axboe:
That sounds good, thanks. Just keep it as incremental patches - first add the helper with something trivial in it, then add further things in subsequent patches. Doing just the malloc first would be a good start, then posix_memalign, etc. Thanks!

axboe:
I'm going to close this one, just open a new pull request with the new approach discussed here.

PR #304, opened 2021-02-22
tests: add helpers.h to harden the tests

ZhiqiangLiu26:
As axboe said, Right now a lot of basic stuff is duplicated,
and it makes the tests bigger than they should be.
Here, we try to add helpers.h which will contains various
utilities that tests could use, then that'd help make tests simpler.

In this patch, we just add one helper io_uring_malloc(), which
will call assert() if allocating memory fails.
We will add more helpers in subsequent patches

### DIFF
--- a/test/helpers.h
+++ b/test/helpers.h
@@ -1,0 +1,21 @@
+/* SPDX-License-Identifier: MIT */
+/*
+ * Description: Helpers for tests.
+ */
+#ifndef LIBURING_HELPERS_H
+#define LIBURING_HELPERS_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/*
+ * Helper for allocating memory in tests.
+ */
+void *io_uring_malloc(size_t size);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif
--- a/test/Makefile
+++ b/test/Makefile
@@ -1,6 +1,8 @@
 prefix ?= /usr
 datadir ?= $(prefix)/share
 
+helpers = helpers.c
+
 INSTALL=install
 
 CPPFLAGS ?=
@@ -132,13 +134,13 @@ test_targets += sq-full-cpp
 endif
 all_targets += sq-full-cpp
 
-all: $(test_targets)
+all: $(test_targets) helpers.o
 
 %: %.c
-	$(QUIET_CC)$(CC) $(CPPFLAGS) $(CFLAGS) -o $@ $< -luring $(XCFLAGS)
+	$(QUIET_CC)$(CC) $(CPPFLAGS) $(CFLAGS) -o $@ $< -luring $(XCFLAGS) ${helpers}
 
 %: %.cc
-	$(QUIET_CXX)$(CXX) $(CPPFLAGS) $(CXXFLAGS) -o $@ $< -luring $(XCFLAGS)
+	$(QUIET_CXX)$(CXX) $(CPPFLAGS) $(CXXFLAGS) -o $@ $< -luring $(XCFLAGS) ${helpers}
 
 test_srcs := \
 	232c93d07b74-test.c \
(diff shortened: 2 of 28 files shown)


ZhiqiangLiu26:
@axboe In this patch, I only add io_uring_malloc() helper in helpers.h file. 
I wonder whether to add only the helpers. h file with static helper functions, or construct the helpers shared library. 
If need to construct the helpers shared library, it seems a little difficult to modify Makefile, and I don't have much experience.
If add only the helpers. h file with static helper functions, gcc will throw warnings (** func defined but not used [-Wunused-function]).
Could you help me?

axboe:
For the include issue, you want them to be marked as ```inline``` as well. But I really do think we need to make this a helpers.c thing and get the Makefile updated for that. It doesn't have to be a shared lib, just linked with every test. You don't have to make it conditional, every test can just link it even if they don't use any of the helpers. Won't really hurt anything.

So would be great if you could take a stab at the Makefile side of things...

ZhiqiangLiu26:
> For the include issue, you want them to be marked as `inline` as well. But I really do think we need to make this a helpers.c thing and get the Makefile updated for that. It doesn't have to be a shared lib, just linked with every test. You don't have to make it conditional, every test can just link it even if they don't use any of the helpers. Won't really hurt anything.
> 
> So would be great if you could take a stab at the Makefile side of things...

thanks for your reply.
I will have a try.

ZhiqiangLiu26:
> For the include issue, you want them to be marked as `inline` as well. But I really do think we need to make this a helpers.c thing and get the Makefile updated for that. It doesn't have to be a shared lib, just linked with every test. You don't have to make it conditional, every test can just link it even if they don't use any of the helpers. Won't really hurt anything.
> 
> So would be great if you could take a stab at the Makefile side of things...


prefix ?= /usr
datadir ?= $(prefix)/share

+helpers = helpers.c

INSTALL=install

CPPFLAGS ?=
@@ -135,10 +137,10 @@ all_targets += sq-full-cpp
all: $(test_targets)

%: %.c
-	$(QUIET_CC)$(CC) $(CPPFLAGS) $(CFLAGS) -o $@ $< -luring $(XCFLAGS)
+	$(QUIET_CC)$(CC) $(CPPFLAGS) $(CFLAGS) -o $@ $< -luring $(XCFLAGS) ${helpers}

%: %.cc
-	$(QUIET_CXX)$(CXX) $(CPPFLAGS) $(CXXFLAGS) -o $@ $< -luring $(XCFLAGS)
+	$(QUIET_CXX)$(CXX) $(CPPFLAGS) $(CXXFLAGS) -o $@ $< -luring $(XCFLAGS) ${helpers}

test_srcs := \
	232c93d07b74-test.c \

axboe:
yeah that looks fine, probably also need:

+all: $(test_targets) helpers.o

to ensure that dependencies get picked up.

test/helpers.h
```diff
@@ -0,0 +1,21 @@
+/* SPDX-License-Identifier: MIT */
+/*
+ * Description: Helpers for tests.
+ */
+#ifndef HELPERS_H
```
axboe, line 5: Should probably make that LIBURING_HELPERS_H just to avoid a name that's too generic and might conflict.

axboe:
With the dep change mentioned and the HELPERS_H changed, I think this can go in.

ZhiqiangLiu26:
> yeah that looks fine, probably also need:
> 
> +all: $(test_targets) helpers.o
> 
> to ensure that dependencies get picked up.

ZhiqiangLiu26:
> With the dep change mentioned and the HELPERS_H changed, I think this can go in.

Thanks for reply.
I have modified the commit as you said.

axboe:
Nice! Thanks for getting this done, helpers.c will only grow in usefulness as time passes and we add more helpers.

YoSTEALTH:
Might be better to rename `io_uring_malloc, io_uring_create_file, ...` to `helper_malloc, helper_create_file, ...` as to avoid confusion. If someone is reading test file its easy to assume these are liburing functions vs helper functions.

axboe:
I agree, it's a bit confusing and overlaps with the liburing namespace. I've renamed them all to be t_ instead, so t_malloc for example.

Issue #310, opened 2021-03-17
Add support for IORING_POLL_ADD_UPDATE


hassila:
Consider adding support for updating to poll operations according to below discussion.

> Or rather that the semantics of IORING_POLL_ADD_MULTI would be that subsequent calls to IORING_POLL_ADD_MULTI with the same user data would replace the previous poll mask in use. Then one could just update away.

I don't like that, because it would mean that any new add would be a lookup as well. If a poll update operation is going to be supported, it should be specifically flagged as such. Eg a ```IORING_POLL_ADD_UPDATE``` flag, for instance, that matches on user_data and re-adds the poll internally based on the new events mask. That would work for both multi and one-shot poll requests.

_Originally posted by @axboe in https://github.com/axboe/liburing/issues/309#issuecomment-801059040_

axboe:
Want to team up on this? I'll update the poll multi branch and add update support, and you run some basic functionality tests on it to ensure that the poll multi works for you, and that the update side does too? That would make my life a lot easier. Bonus points if the functional tests are stand-alone so I can steal them for the regression suite as well :-)

hassila:
Sure, I'd be happy to help out - may need some pointer on where to find info on how to build the kernel pulling in those changes (I'm currently running on Ubuntu 20.04 with a 5.11 kernel put on top for uring). I'm basically testing this out in context of implementing https://github.com/apple/swift-nio/issues/1761 but would be happy to write some stand-alone C basic functionality tests too.

axboe:
https://git.kernel.dk/cgit/linux-block/log/?h=poll-multiple

This has been updated with the patch adding support for poll modifications. They may be done on either a one-shot of multi-shot poll request. See that top commit for a description.

In terms of testing, I'd just poll that branch, build, and boot it. It'll be based on 5.12-rc3 with various things already queued up for 5.13.

hassila:
That's great, I will have a look!

axboe:
FWIW, I did run a simple test case:

1) arm poll request with POLLOUT for the read side of a pipe
2) Check it doesn't complete when data is written to the other side of the pipe
3) Then issue update request with POLLIN set instead
4) Notice that update request completes with res == 0, and original poll request now completes with POLLIN available

That's the extent of testing I did on that patch, JFYI ;-)

axboe:
Just a heads up that I tested multi shot too now, and there's a mask issue so it'll end up turning into a one-shot after that. Fixing that now.

[...]

PR #311, opened 2021-03-18
First simple test for multi-shot fd updates for #310

hassila:
For future merging - this requires the io_uring header from https://git.kernel.dk/cgit/linux-block/log/?h=poll-multiple to be integrated first with liburing before merging as it uses IORING_POLL_ADD_MULTI.


hassila:

PR #314, opened 2021-03-20
Test for #310 ping-pong which receives spurious double poll events

hassila:
For future possible merging when poll multi shot + modifications are integrated.


hassila:

PR #315, opened 2021-03-20
#310 test for detecting NOHUP when other side closes socket

hassila:
For future merging when multi shot polls have been integrated.

Test that have a client/server where the client write x times over a socket, then closes it. The server should receive a POLLHUP at the end (currently does not, just receive 0 bytes read and POLLIN).


hassila:

PR #316, opened 2021-03-21
Added test for #310, checking expected failures for ENOENT, EBADF and EINVAL

hassila:
For future merging when multishot polls are integrated.

One more test for #310, this one tests that we can trigger ENOENT, EBADF and EINVAL as expected.


hassila:

PR #335, opened 2021-04-29
examples/kdigest: add AF_ALG hash example

ddiss:
When built with CONFIG_CRYPTO_USER_API_HASH enabled, Linux exposes a
socket based API for hashing data. When coupled with uring
IOSQE_IO_LINK, file hashing can be done in quite an efficient manner,
as demonstrated in this link-cp.c based example.

Signed-off-by: David Disseldorp <ddiss@suse.de>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@
 /examples/io_uring-test
 /examples/link-cp
 /examples/ucontext-cp
+/examples/kdigest
 
 /test/232c93d07b74-test
 /test/35fa71a030ca-test
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -10,7 +10,7 @@ ifneq ($(MAKECMDGOALS),clean)
 include ../config-host.mak
 endif
 
-all_targets += io_uring-test io_uring-cp link-cp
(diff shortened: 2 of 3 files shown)


chrisvest:
Looks like the linked pairs be processed out of order, so the hash result is non-deterministic, no?

ddiss:
> Looks like the linked pairs be processed out of order, so the hash result is non-deterministic, no?

Yes,  you're right. I'll close this while I rework it. Setting `QD` to one would be the most obvious fix but will likely hurt performance...

ddiss:
I think flagging `IOSQE_IO_DRAIN` on the AF_ALG send I/O would be the simplest fix for hash ordering here... IIUC it'll still permit AF_ALG send and next-offset file read I/Os to run concurrently, but everything else will move in lock step.

ddiss:
> I think flagging `IOSQE_IO_DRAIN` on the AF_ALG send I/O would be the simplest fix for hash ordering here... IIUC it'll still permit AF_ALG send and next-offset file read I/Os to run concurrently, but everything else will move in lock step.

Looks like my force push has blocked reopening here. Will submit this change as a new PR.

PR #336, opened 2021-04-29
examples/kdigest: add AF_ALG hash example

ddiss:
When built with CONFIG_CRYPTO_USER_API_HASH enabled, Linux exposes a
socket based API for hashing data. When coupled with uring
IOSQE_IO_LINK, file hashing can be done in quite an efficient manner,
as demonstrated in this link-cp.c based example.

Signed-off-by: David Disseldorp <ddiss@suse.de>

### DIFF
--- a/examples/kdigest.c
+++ b/examples/kdigest.c
@@ -1,0 +1,305 @@
+/* SPDX-License-Identifier: MIT */
+/*
+ * Proof-of-concept for doing file digests using the kernel's AF_ALG API.
+ * Needs a bit of error handling.
+ */
+#include <stdio.h>
+#include <fcntl.h>
+#include <string.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <assert.h>
+#include <errno.h>
+#include <inttypes.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <sys/ioctl.h>
+#include <linux/if_alg.h>
+#include "liburing.h"
+
+#define QD	64
+#define BS	(32*1024)
+
+enum req_state {
+	IO_INIT = 0,
+	IO_READ,
+	IO_READ_COMPLETE,
+	IO_WRITE,
+	IO_WRITE_COMPLETE,
+};
+
+struct req {
+	off_t offset;
+	enum req_state state;
+	struct iovec iov;
+};
+
+struct kdigest {
+	struct io_uring ring;
+	struct req reqs[QD];
+	/* heap allocated, aligned QD*BS buffer */
+	uint8_t *bufs;
+};
+
+static int infd, outfd;
+
+static int get_file_size(int fd, size_t *size)
+{
+	struct stat st;
+
+	if (fstat(fd, &st) < 0)
+		return -1;
+	if (S_ISREG(st.st_mode)) {
+		*size = st.st_size;
+	} else if (S_ISBLK(st.st_mode)) {
+		unsigned long long bytes;
+
+		if (ioctl(fd, BLKGETSIZE64, &bytes) != 0)
+			return -1;
+
+		*size = bytes;
+	} else {
+		return -1;
+	}
+
+	return 0;
+}
+
+static int digest_file(struct kdigest *kdigest, size_t insize)
+{
+	struct io_uring *ring = &kdigest->ring;
+	off_t read_off = 0;
+	size_t outsize = insize;
+	int read_idx = 0, write_idx = 0, inflight = 0;
+
+	while (outsize) {
+		int to_wait;
+		struct req *req;
+		struct io_uring_sqe *sqe;
+		int had_inflight = inflight;
+
+		/* Queue up any possible writes. Link flag ensures ordering. */
+		sqe = NULL;
+		while (kdigest->reqs[write_idx].state == IO_READ_COMPLETE) {
+			if (sqe)
+				sqe->flags |= IOSQE_IO_LINK;
+
+			req = &kdigest->reqs[write_idx];
+			req->state = IO_WRITE;
+			sqe = io_uring_get_sqe(ring);
+			io_uring_prep_send(sqe, outfd, req->iov.iov_base,
+					   req->iov.iov_len, MSG_MORE);
+			io_uring_sqe_set_data(sqe, req);
+			inflight++;
+
+			write_idx = (write_idx + 1) % QD;
+		}
+
+		/* Queue up any reads. Completions may arrive out of order. */
+		while (insize && (kdigest->reqs[read_idx].state == IO_INIT
+		    || kdigest->reqs[read_idx].state == IO_WRITE_COMPLETE)) {
+			size_t this_size = (insize < BS ? insize : BS);
+
+			req = &kdigest->reqs[read_idx];
+			req->state = IO_READ;
+			req->offset = read_off;
+			req->iov.iov_base = &kdigest->bufs[read_idx * BS];
+			req->iov.iov_len = this_size;
+
+			sqe = io_uring_get_sqe(ring);
+			io_uring_prep_readv(sqe, infd, &req->iov, 1, read_off);
+			io_uring_sqe_set_data(sqe, req);
+
+			read_off += this_size;
+			insize -= this_size;
+			inflight++;
+
+			read_idx = (read_idx + 1) % QD;
+		}
+
+		if (had_inflight != inflight) {
+			assert(inflight > had_inflight);
+			if (io_uring_submit(ring) < 0)
+				return 1;
+		}
+
+		/* wait for about half queue completion before resubmit */
+		for (to_wait = (inflight >> 1) | 1; to_wait; to_wait--) {
+			struct io_uring_cqe *cqe;
+			int ret;
+
+			ret = io_uring_wait_cqe(ring, &cqe);
+			if (ret < 0) {
+				fprintf(stderr, "wait cqe: %s\n",
+					strerror(-ret));
+				return 1;
+			}
+
+			req = io_uring_cqe_get_data(cqe);
+			assert(req->state == IO_READ || req->state == IO_WRITE);
+			ret = cqe->res;
+			io_uring_cqe_seen(ring, cqe);
+			if (ret < 0) {
+				if (ret == -ECANCELED && req->state == IO_READ) {
+					fprintf(stderr, "canceled read@%lld\n",
+						(long long)req->offset);
+					sqe = io_uring_get_sqe(ring);
+					io_uring_prep_readv(sqe, infd,
+						&req->iov, 1, req->offset);
+					io_uring_sqe_set_data(sqe, req);
+					if (io_uring_submit(ring) < 0)
+						return 1;
+					continue;
+				} else {
+					fprintf(stderr, "cqe error: %s\n",
+						strerror(-ret));
+					return 1;
+				}
+			}
+
+			inflight--;
+			req->state++;
+
+			if (req->state == IO_WRITE_COMPLETE)
+				outsize -= req->iov.iov_len;
+		}
+	}
+	assert(!inflight);
+
+	return 0;
+}
+
+static int get_result(struct io_uring *ring, const char *alg, const char *file)
+{
+	struct io_uring_sqe *sqe;
+	struct io_uring_cqe *cqe;
+	int i, ret;
+	/* buffer must be large enough to carry longest hash result */
+	uint8_t buf[4096];
+
+	sqe = io_uring_get_sqe(ring);
+	memset(buf, 0, sizeof(buf));
+	io_uring_prep_read(sqe, outfd, buf, sizeof(buf), 0);
+	if (io_uring_submit(ring) < 0)
+		return 1;
+
+	ret = io_uring_wait_cqe(ring, &cqe);
+	if (ret < 0) {
+		fprintf(stderr, "wait cqe: %s\n", strerror(-ret));
+		return 1;
+	}
+
+	if (cqe->res < 0 || cqe->res > sizeof(buf)) {
+		fprintf(stderr, "cqe error: %s\n", strerror(-cqe->res));
+		ret = 1;
+	} else {
+		fprintf(stdout, "uring %s(%s) returned(len=%u): ",
+			alg, file, cqe->res);
+		for (i = 0; i < cqe->res; i++)
+			fprintf(stdout, "%02x", buf[i]);
+		putc('\n', stdout);
+		ret = 0;
+	}
+
+	io_uring_cqe_seen(ring, cqe);
+	return ret;
+}
+
+int main(int argc, char *argv[])
+{
+	const char *alg;
+	const char *infile;
+	size_t alg_len;
+	struct sockaddr_alg sa = {
+		.salg_family = AF_ALG,
+		.salg_type = "hash",
+	};
+	int sfd = -1;
+	size_t insize;
+	int ret;
+	struct kdigest kdigest = {};
+
+	if (argc < 3) {
+		fprintf(stderr, "%s: algorithm infile\n", argv[0]);
+		return 1;
+	}
+
+	alg = argv[1];
+	infile = argv[2];
+	alg_len = strlen(alg);
+	if (alg_len >= sizeof(sa.salg_name)) {
+		fprintf(stderr, "algorithm name too long\n");
+		return 1;
+	}
+	/* +1 for null terminator */
+	memcpy(sa.salg_name, alg, alg_len + 1);
+
+	infd = open(infile, O_RDONLY);
+	if (infd < 0) {
+		perror("open infile");
+		return 1;
+	}
+
+	sfd = socket(AF_ALG, SOCK_SEQPACKET, 0);
+	if (sfd < 0) {
+		if (errno == EAFNOSUPPORT)
+			fprintf(stderr, "kernel AF_ALG support not available. "
+				"CONFIG_CRYPTO_USER_API_HASH required.\n");
+		else
+			perror("AF_ALG socket");
+		return 1;
+	}
+
+	if (bind(sfd, (struct sockaddr *)&sa, sizeof(sa)) < 0) {
+		if (errno == ENOENT)
+			fprintf(stderr, "AF_ALG bind(%s): hash not available. "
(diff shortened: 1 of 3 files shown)


ddiss:
v1 was submitted via https://github.com/axboe/liburing/pull/335 .

examples/kdigest.c
```diff
@@ -0,0 +1,284 @@
...
+	struct io_uring_sqe *sqe;
+	struct io_uring_cqe *cqe;
+	int i, ret;
+	/* buffer must be large enough to carry longest hash result */
+	uint8_t buf[4096];
+
+	sqe = io_uring_get_sqe(ring);
+	memset(buf, 0, sizeof(buf));
+	io_uring_prep_read(sqe, outfd, buf, sizeof(buf), 0);
+	io_uring_submit(ring);
```
isilence, line 179: Better to check errors, especially since it's an example

examples/kdigest.c
```diff
@@ -0,0 +1,284 @@
...
+	io_uring_prep_readv(sqe, infd, &data->iov, 1, offset);
+	sqe->flags |= IOSQE_IO_LINK;
+	io_uring_sqe_set_data(sqe, data);
+
+	sqe = io_uring_get_sqe(ring);
+	/*
+	 * data must be sent to the hasher in file order. IOSQE_IO_DRAIN is
+	 * needed here to avoid concurrent / out of order send I/Os.
+	 */
+	sqe->flags |= IOSQE_IO_DRAIN;
```
isilence, line 90: Haven't looked through the code, but drain is very far from anything performant, doesn't sound like it goes well with the idea of performant hashing

examples/kdigest.c
ddiss: Okay, will add this. FWIW this is based on the link-cp example which is similarly reckless.

isilence: Hah, that's not great indeed

examples/kdigest.c
ddiss: Yes, it's not ideal. One option I'd considered is batching up a bunch of concurrent read requests and then `IOSQE_IO_LINK` chaining subsequent hash send requests for the batch. A drain would still be need to on the first send, so I'm not sure whether the extra complexity would be worth it. I'm open to other ideas :-)

I've done some basic benchmarks of this series against openssl and don't see much of a difference, but want to confirm that I'm not comparing optimized vs non-optimized numbers before posting numbers. Thanks for the feedback.

ddiss:
Changes since previous version:
- add some more crude error checks
- print newline after hash

examples/kdigest.c
ddiss: FWIW, here are some rough benchmark results that I ran on my laptop (VM: 2 core, 2G RAM, v5.12 kernel, openssl-1_1-1.1.1d-lp152.7.15.1.x86_64):
### **kdigest**
```
\ SIZE      512 bytes               4096              65536            1048576           16777216           33554432
HASH
   md5 | 0.000650+- 5.26% | 0.000646+- 4.64% | 0.000815+- 3.15% | 0.002965+- 0.95% | 0.033185+- 0.13% | 0.066930+- 2.14% | 
  sha1 | 0.000627+- 3.40% | 0.000614+- 6.30% | 0.000799+- 3.80% | 0.003165+- 1.14% | 0.036577+- 0.23% | 0.075130+- 2.34% | 
sha224 | 0.000619+- 4.49% | 0.000655+- 3.85% | 0.000989+- 3.82% | 0.005848+- 0.65% | 0.078375+- 0.37% | 0.156153+- 0.21% | 
sha256 | 0.000616+- 2.10% | 0.000662+- 2.92% | 0.001017+- 3.51% | 0.005794+- 0.44% | 0.078351+- 0.38% | 0.156910+- 1.28% | 
sha384 | 0.000607+- 4.26% | 0.000661+- 4.26% | 0.000871+- 3.68% | 0.004977+- 0.81% | 0.065744+- 0.29% | 0.130322+- 0.30% | 
sha512 | 0.000640+- 5.56% | 0.000647+- 6.58% | 0.001077+- 5.46% | 0.004986+- 0.74% | 0.065631+- 0.14% | 0.130303+- 0.33% | 
```

### **openssl**
```
\ SIZE      512 bytes               4096              65536            1048576           16777216           33554432
HASH
   md5 | 0.003150+-11.17% | 0.003123+- 9.36% | 0.003344+-15.23% | 0.005256+- 5.92% | 0.034233+- 1.56% | 0.068180+- 2.44% | 
  sha1 | 0.002534+- 1.64% | 0.002554+- 1.51% | 0.002662+- 2.49% | 0.004902+- 2.75% | 0.036007+- 1.06% | 0.069026+- 0.57% | 
sha224 | 0.002512+- 1.42% | 0.002569+- 2.12% | 0.002868+- 1.17% | 0.006807+- 0.67% | 0.070918+- 0.40% | 0.138509+- 0.06% | 
sha256 | 0.002526+- 2.10% | 0.002541+- 1.52% | 0.002821+- 2.83% | 0.006830+- 0.46% | 0.070984+- 0.45% | 0.153090+- 2.97% | 
sha384 | 0.002513+- 1.39% | 0.002501+- 1.41% | 0.002746+- 2.96% | 0.005408+- 0.58% | 0.047937+- 0.04% | 0.093216+- 0.40% | 
sha512 | 0.002534+- 2.18% | 0.002519+- 1.91% | 0.002932+- 4.21% | 0.005413+- 0.49% | 0.048250+- 0.47% | 0.093774+- 0.25% | 
```

Benchmark script:
```
...
for size in $((32 * 1024 * 1024)) $((16 * 1024 * 1024)) $((1024 * 1024)) \
            $((64 * 1024)) $((4 * 1024)) 512; do
        echo "==== hashing file of size $size ===="
        for i in md5 sha1 sha224 sha256 sha384 sha512; do
                # prime cache
                cat "${size}.data" > /dev/null
                perf stat --null -r 5 --table openssl "$i" "${size}.data" >/dev/null 2>openssl.${size}.${i}.perf
                perf stat --null -r 5 --table liburing/examples/kdigest "$i" "${size}.data" >/dev/null 2>kdigest.${size}.${i}.perf
        done
done
```

I did some benchmarks using `splice`(x2) instead of `readv`+`send`, but didn't see an improvement in performance (it actually dropped a little). Given that this is supposed to just be a simple example / poc I'd prefer to keep things simple.

ddiss:
@isilence any further thoughts on this patchset, or should I drop it?
FWIW, basic benchmark results can be found at https://github.com/axboe/liburing/pull/336#discussion_r625789647

ddiss:
> @isilence any further thoughts on this patch, or should I drop it?
> FWIW, basic benchmark results can be found at [#336 (comment)](https://github.com/axboe/liburing/pull/336#discussion_r625789647)

@axboe feel free to close if there's no interest in this change.

isilence:
I don't really mind it get merged if it's sane enough, but apparently nobody just have time to take a look. I'd rather assume it to stay until then

examples/kdigest.c
```diff
@@ -0,0 +1,292 @@
...
+
+	io_uring_cqe_seen(ring, cqe);
+	return ret;
+}
+
+int main(int argc, char *argv[])
+{
+	struct io_uring ring;
+	const char *alg = argv[1];
+	const char *infile = argv[2];
```
isilence, line 211: Not before argc check

examples/kdigest.c
```diff
@@ -0,0 +1,292 @@
...
+struct io_data {
+	size_t offset;
+	int index;
+	struct iovec iov;
+};
+
+static int infd, outfd;
+static unsigned inflight;
+
+static int setup_context(unsigned entries, struct io_uring *ring)
```
isilence, line 32: Why the function does even exist but not inlined?

examples/kdigest.c
```diff
@@ -0,0 +1,292 @@
...
+			return -1;
+
+		*size = bytes;
+		return 0;
+	}
+
+	return -1;
+}
+
+static void queue_rw_pair(struct io_uring *ring, off_t size, off_t offset)
```
isilence, line 67: size_t?

examples/kdigest.c
```diff
@@ -0,0 +1,292 @@
...
+	sqe->flags |= IOSQE_IO_LINK;
+	io_uring_sqe_set_data(sqe, data);
+
+	sqe = io_uring_get_sqe(ring);
+	/*
+	 * data must be sent to the hasher in file order. IOSQE_IO_DRAIN is
+	 * needed here to avoid concurrent / out of order send I/Os.
+	 */
+	sqe->flags |= IOSQE_IO_DRAIN;
+	io_uring_prep_send(sqe, outfd, ptr, size, MSG_MORE);
```
isilence, line 92: So, it clears up back to 0 the drain flag you set just above, so you had it disabled all the time

examples/kdigest.c
```diff
@@ -0,0 +1,292 @@
...
+		if (has_inflight != inflight) {
+			if (io_uring_submit(ring) < 0)
+				return 1;
+		}
+
+		if (insize)
+			depth = QD;
+		else
+			depth = 1;
+		while (inflight >= depth) {
```
isilence, line 153: ```while (inflight)``` ?

isilence: Or maybe even ```to_wait = insize ? QD / 2 : 1``` to batch submissions

examples/kdigest.c
isilence: And I assume it will go down once you fix it, because you basically serialise all your QD=64.
Still, I'd suggest to do something smarter instead of draining

examples/kdigest.c
isilence: Is that file size? For large sizes openssl works even better

examples/kdigest.c
```diff
@@ -0,0 +1,292 @@
...
+	return -1;
+}
+
+static void queue_rw_pair(struct io_uring *ring, off_t size, off_t offset)
+{
+	struct io_uring_sqe *sqe;
+	struct io_data *data;
+	void *ptr;
+
+	ptr = malloc(size + sizeof(*data));
```
isilence, line 73: How about preallocating it? +align

examples/kdigest.c
ddiss: Will fix in next round.

examples/kdigest.c
ddiss: I think this is carryover from link-cp. I'll look at simplifying this next round.

examples/kdigest.c
ddiss: ack

examples/kdigest.c
ddiss: Aside from LINKing across multiple batches, I struggle to find an option for providing the required AF_ALG write ordering without a DRAIN - reads can run concurrently and out of order, but writes must proceed in file offset order.

examples/kdigest.c
ddiss: Yes that makes sense. I did play with `QD / 2` here for the splice variant but as mentioned, it didn't show any clear performance improvement.

examples/kdigest.c
ddiss: ack, will do this in the next round.

examples/kdigest.c
isilence: With drain you just get it serialised, so no better than synchronous read/write.

E.g. forget about links, run reads as concurrent as you can (but with QD as you don't have infinite amount of buffers), save the fact of completion of chunks (e.g. with a bitmap) and write only when it's sequentially ready. May require some tweaking with buffer management.

isilence: Like

```C
cqe = get_completion();
set_bit(bitmap, cqe->off / REQ_SIZE);
if (test_bit(btimap, current_bit)) {
    send_write_sqe();
    current_bit++;
}
```

isilence: If tuned good enough should be pretty similar for large files

ddiss: will give it a try. Thanks a lot for the feedback

isilence: Actually, performance may get worse in the ideal program, depends on time required to hash compared to read... Though, considering that openssl does have a very similar runtime for large files, it may be relatively negligible for your case.

Btw, as you serialise over bitmaps -- you can link together requests writing to the hash socket as much as is available. Or send one big in the first place (setting the right iov).

axboe:
@ddiss Were you planning on respinning this change?

ddiss:
> @ddiss Were you planning on respinning this change?

Thanks for the ping. Yes, the respin is still on my todo list, but some other things got in the way. I'm hoping to return to this in a few weeks. I'll close here for now and reopen when ready.

ddiss:
I finally got a chance to revisit this, so will post an update here, hopefully addressing Pavel's feedback.

ddiss:
changes since 2021:
- rewrite dispatch logic to submit reads in parallel alongside linked writes I/Os
- use a simple read/write index + state for tracking sequential submission
- preallocate aligned I/O buffers

axboe:
Thanks for picking this up again! Can you address the CI failures? Looks like it's just basic complaints on things that should be static. I think this would be a good example to add.

ddiss:
changes to address CI failures:
- drop unused debug macro / statements
- cast `off_t` for `fprintf()`

axboe:
Rather than use ```IOSQE_IO_LINK```, an improvement here may be to use ring provided buffers and bundle send to send data. That should be more efficient, at least on kernels that support both of those features.

axboe:
https://github.com/axboe/liburing/commit/73a70031805e2d91a4d4794cbbcc8e05ed4bf847

That adds bundled sends. Didn't test it much, but seems to work.

axboe:
Ran a quick test on an amd box, and I ranges from ~5x faster at the lower sizes, to being even for 16MB/32MB. Ran your test on it. So whether that helped or not, not sure, but at least it's no longer slower for bigger sizes and it's still considerably faster for sizes that people would actually use.

ddiss:
Thanks for the follow up changes. FWIW, I did play around with merging buffers for the send, but didn't see much perf improvement for the added complexity. Bundling looks good.

axboe:
Right, with bundles it's just as easy as using links, at least. But I like having both in the example, as it shows how to use either one.

ddiss:
In case anyone is interested, below are some rough kdigest ( c90e5a6 ) perf measurements:

## kdigest-c90e5a6
```
  \ FILE
   \SIZE    512 bytes     |       4096       |      65536       |     1048576      |     16777216     |     33554432     |
HASH\__|__________________|__________________|__________________|__________________|__________________|__________________|
   md5 | 0.0010984+-1.44% | 0.0008265+-2.65% | 0.0009556+-1.52% | 0.0024098+-0.35% | 0.0266533+-0.06% | 0.0521402+-0.19% |
  sha1 | 0.0011012+-1.59% | 0.0009430+-2.21% | 0.0009173+-1.89% | 0.0019097+-1.29% | 0.0186466+-0.18% | 0.0361400+-0.13% |
sha224 | 0.0010983+-1.29% | 0.0010970+-1.72% | 0.0010350+-1.32% | 0.0036209+-0.42% | 0.0425834+-0.06% | 0.0841893+-0.06% |
sha256 | 0.0010996+-1.34% | 0.0011159+-1.74% | 0.0010299+-1.09% | 0.0036085+-0.40% | 0.0426466+-0.12% | 0.0840805+-0.03% |
sha384 | 0.0011094+-1.58% | 0.0011204+-1.48% | 0.0009555+-2.96% | 0.0027775+-0.58% | 0.0296736+-0.27% |  0.058271+-0.27% |
sha512 | 0.0010909+-1.71% | 0.0010763+-3.21% | 0.0009746+-1.77% | 0.0027719+-0.74% | 0.0297744+-0.26% |  0.058239+-0.25% |
```

## openssl-3.1.4-3.2
```
  \ FILE
   \SIZE    512 bytes     |       4096       |      65536       |     1048576      |     16777216     |     33554432     |
HASH\__|__________________|__________________|__________________|__________________|__________________|__________________|
   md5 | 0.0039263+-0.81% | 0.0029834+-1.69% | 0.0030833+-1.28% | 0.0044167+-0.87% | 0.0286969+-0.20% | 0.0536009+-0.16% |
  sha1 | 0.0039302+-1.16% | 0.0029809+-1.62% | 0.0030169+-0.99% | 0.0040051+-1.04% | 0.0220672+-0.29% | 0.0414711+-0.19% |
sha224 | 0.0039211+-0.99% | 0.0039417+-0.95% | 0.0031360+-1.43% | 0.0055392+-0.69% | 0.0408525+-0.18% |  0.078564+-0.20% |
sha256 | 0.0039277+-0.60% | 0.0039653+-0.97% | 0.0031659+-1.26% | 0.0055284+-0.76% | 0.0408774+-0.11% | 0.0788206+-0.10% |
sha384 | 0.0039370+-1.13% | 0.0039494+-0.87% | 0.0030840+-1.26% | 0.0047742+-0.81% |  0.029442+-0.35% |  0.056091+-0.24% |
sha512 | 0.0039456+-1.02% | 0.0039779+-1.09% | 0.0030739+-1.26% | 0.0047586+-0.55% | 0.0294435+-0.20% |  0.056350+-0.31% |
```

Benchmark script:
```shell
for size in $((32 * 1024 * 1024)) $((16 * 1024 * 1024)) $((1024 * 1024)) \
            $((64 * 1024)) $((4 * 1024)) 512; do
    dd if=/dev/urandom of="${size}.data" bs="$size" count=1 || break
    echo "==== hashing file of size $size ===="
    for i in md5 sha1 sha224 sha256 sha384 sha512; do
        # prime cache
        cat "${size}.data" > /dev/null
        perf stat --null -r 5 --table openssl "$i" "${size}.data" >/dev/null 2>openssl.${size}.${i}.perf
        perf stat --null -r 5 --table ~/liburing/examples/kdigest "$i" "${size}.data" >/dev/null 2>kdigest.${size}.${i}.perf
    done
done
```

Host details:
```
    Kernel: openSUSE Tumbleweed 6.11.0-1-default
    CPU: Intel(R) Xeon(R) CPU E3-1260L v5 @ 2.90GHz
    Thread(s) per core:   2
    Core(s) per socket:   4
    Socket(s):            1
    RAM: 64GB
```

Issue #337, opened 2021-05-04
io_uring_queue_exit returns void, man page says it returns int.


le-migou:
Which one should it be?

isilence:
https://github.com/axboe/liburing/blob/7ff4feed7e93e42fb7cec689f7131535f1a128ca/man/io_uring_queue_exit.3#L22-L23

https://github.com/axboe/liburing/blob/7ff4feed7e93e42fb7cec689f7131535f1a128ca/src/include/liburing.h#L113

Please specify where man says so.

le-migou:
In the synopsis, line 13:

    .BI "int io_uring_queue_exit(struct io_uring *" ring );

isilence:
> In the synopsis, line 13:
> 
> ```
> .BI "int io_uring_queue_exit(struct io_uring *" ring );
> ```

Got it, should be corrected. The function returns nothing since the very first day.

isilence:
Want to send a patch fixing the description?

isilence:
And `"`s are set strangely in that line as well

le-migou:
Here it is.

Le 04/05/2021 à 12:46, Pavel Begunkov a écrit :

le-migou:
Sorry I don't know how to join a patch to a comment. I'll make a PR.

isilence:
It's either a patch to `io-uring@vger.kernel.org`, or Jens takes PRs

PR #338, opened 2021-05-04
Fix typo in man io_uring_queue_exit

le-migou:
Regarding #337 .


isilence:
there are certain guidelines, though not too strict for liburing

1) please add a commit message, not only title
2) add signed-off tag, see any commit, e.g. https://github.com/axboe/liburing/commit/7ff4feed7e93e42fb7cec689f7131535f1a128ca
Looks like `Signed-off-by: Pavel Begunkov <my_email@example.com>`
3) preferable to add "man:" to the beginning of the commit title

PR #362, opened 2021-06-12
test/file-register: fix heap overflow read on test_basic

ammarfaizi2:
When `test_basic` is called with `fail == 1`, the allocated array for
it is just 10, but the close_files is always called with 100.


axboe:
Can you rebase this on top of master? Seems to be some unrelated history in yours.

axboe:
Nevermind, that was me being an idiot. Thanks!

Issue #378, opened 2021-07-15
Potential error(e.g., deadlock) due to the unreleased lock &lock


anson-lo:
Hi, developers:
       It seems the lock `$lock `is not released if the control flow goes to the` break` statement. The fix is to add the unlock statement before the `break`. Thank you for your checking! It is in the test folder but still worth fixing for higher quality.

The suggested fix is to add `pthread_mutex_unlock(&lock);` before breaking and returning.

https://github.com/axboe/liburing/blob/a209abe65befe603311b6fcd861e14d28e3e2f2d/test/ring-leak2.c#L194
https://github.com/axboe/liburing/blob/a209abe65befe603311b6fcd861e14d28e3e2f2d/test/ring-leak2.c#L198

The relevant code is shown here.

```
static void *client_thread(void *arg)
{
	...; // while loop
			if (type == READ) {
				pthread_mutex_lock(&lock); //the lock is here

				if (cqe->res <= 0) {
					// connection closed or error
					shutdown(conn_i.fd, SHUT_RDWR);
				} else {
					break;   //if it is breaked, the lock is not released
				}
				add_socket_pollin(&ring, conn_i.fd);
				pthread_mutex_unlock(&lock);
			} ...;
		}
              ...;
	}
}

```
Best,

isilence:
Hey @anson-lo, looks you did a good job looking for problems in tests. Do you want to send patches or a PR? Would be appreciated.

anson-lo:
@isilence Thank you for your reply. Sure! I would send the patches soon.

anson-lo:
The patch for this issue. [myfixes.patch.zip](https://github.com/axboe/liburing/files/6827452/myfixes.patch.zip)

```
diff -Naur liburing/test/ring-leak2.c liburing-fix/test/ring-leak2.c
--- liburing/test/ring-leak2.c	2021-07-16 10:35:19.000000000 +0800
+++ liburing-fix/test/ring-leak2.c	2021-07-16 10:37:56.000000000 +0800
@@ -197,6 +197,7 @@
 					// connection closed or error
 					shutdown(conn_i.fd, SHUT_RDWR);
 				} else {
+					pthread_mutex_unlock(&lock);
 					break;
 				}
 				add_socket_pollin(&ring, conn_i.fd);
```

axboe:
Can you send that as a real patch or a pull request? Note the usual format of commits in the tree.

anson-lo:
[...]

PR #384, opened 2021-07-18
fix a potential bug due to the unreleased lock for issue#378

anson-lo:
Patch for the issue https://github.com/axboe/liburing/issues/378.

### DIFF
--- a/test/ring-leak2.c
+++ b/test/ring-leak2.c
@@ -197,7 +197,7 @@ static void *client_thread(void *arg)
 					// connection closed or error
 					shutdown(conn_i.fd, SHUT_RDWR);
 				} else {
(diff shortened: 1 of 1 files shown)


axboe:
Please squash the commits, no point in having a fix and a fix for a fix commit in the tree. Also please heed the advice provided on formatting your commit correctly. First line is title, empty line, description, empty line, then a signed-off-by with your name and email. It's good practice to just browse the last handfuls of commits in a repo when submitting patches. A well managed repository has a consistent style to it. Here's a recent example:

https://git.kernel.dk/cgit/liburing/commit/?id=f652f3c91806bd6efb10cb8e8a68bfbe58843e98

PR #389, opened 2021-07-22
Fix a bug due to the unreleased lock before function returns

anson-lo:
Fix a bug due to the unreleased lock before the function returns by inserting unlock statement before breaking.

Signed-off-by: Anson Lo ycaibb@gmail.com


axboe:
Pulled, thanks.

PR #394, opened 2021-07-26
test/io_uring_enter.c: set to_submit to 0

seunghunee:
`to_submit` argument probably should be 0

Signed-off-by: 신승훈 <seunghunee@gmail.com>


axboe:
Doesn't matter in practice as the ring is empty, but yes it doesn't make sense to use '1' here.

PR #395, opened 2021-07-27
Clarify information about error results

MiKom:
One could (and did) interpret that cqe->res mathes the return *value* of preadv2 and pwritev2.
Clarify that it matches the errno values used by these two.

### DIFF
--- a/man/io_uring_enter.2
+++ b/man/io_uring_enter.2
@@ -944,7 +944,7 @@ They are described in section
 .B CQE ERRORS.
 .PP
 For read and write opcodes, the
-return values match those documented in the
+return values match errno values documented in the
 .BR preadv2 (2)
 and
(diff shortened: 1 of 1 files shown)


axboe:
Looks good! Might be worth making it fully complete and mention that cqe->res is -errno, not just for read/write operation but in general for any request type that is a mirror of the equivalent system call?

MiKom:
Thanks for the tips! I should have some time to look deeper during the weekend.

axboe:
I can pull this one for now, just send an updated one when you have time. Thanks!

axboe:
Pulled it, and added a commit on top. Hopefully this will do it... Change in question:

https://git.kernel.dk/cgit/liburing/commit/?id=4c23254cc770505816a307ed8c4f17f9a4aaefb9

PR #410, opened 2021-08-13
correct syscall NR in mips

yangbonis:
I want to use iouring in mips, find some problem include kernel and library. 

Signed-off-by: wangyangbo <wangyangbo@uniontech.com>


axboe:
Thanks!

PR #416, opened 2021-08-19
examples: disable ucontext-cp for elbrus (e2k) architecture

pechenkin:
Elbrus 2000 (aka e2k) is a 64-bit little-endian architecture. For several reasons it use makecontext_e2k(ctx, ...) with different semantic instead of makecontext(ctx, ...).

### DIFF
--- a/configure
+++ b/configure
@@ -327,6 +327,7 @@ int main(int argc, char **argv)
 {
   ucontext_t ctx;
   getcontext(&ctx);
+  makecontext(&ctx, 0, 0);
   return 0;
 }
 EOF


ammarfaizi2:
@pechenkin, can you add `Signed-off-by` tag at the end of your commit message?

As far as I can see, Jens requires commit to contain `Signed-off-by` tag. The format looks like this:
```
Signed-off-by: Your Name <youremail@domain.com>
```
See also the past commit for example: https://github.com/axboe/liburing/commit/72a943982615b655d0cfc08f882ab91079f3287e

pechenkin:
@ammarfaizi2 I fixed it. I thought it was reviewer`s sign.

PR #425, opened 2021-09-07
Test suite shellcheck fixes

guillemj:
I noticed the "! -z" oddness, and after fixing it decided to have a go with shellcheck, which resulted in this patch series.

I'm also thinking this would be nice to run as part of the repo's CI, which I notice is marked as failing, and I assume travis is not working either given their recent policy change. If you have no issue with this, I'd probably work tomorrow on switching these to use github actions?

### DIFF
--- a/test/runtests-loop.sh
+++ b/test/runtests-loop.sh
@@ -1,10 +1,10 @@
 #!/bin/bash
 
-TESTS="$@"
+TESTS=("$@")
 ITER=0
 
 while true; do
-	./runtests.sh "$TESTS"
+	./runtests.sh "${TESTS[@]}"
 	RET="$?"
 	if [ "${RET}" -ne 0 ]; then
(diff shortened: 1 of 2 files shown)


ammarfaizi2:
> I'm also thinking this would be nice to run as part of the repo's CI, which I notice is marked as failing, and I assume travis is not working either given their recent policy change. If you have no issue with this, I'd probably work tomorrow on switching these to use github actions?

The failing appveyor is annoying to see indeed. But speaking of CI, liburing is some kind of kernel test, which requires us to install the latest kernel. Can GitHub actions do that?

ammarfaizi2 (approved):

axboe:
CI does need some love, it's pretty much just disabled. There's probably limited utility in actually running the tests, however building the lib and tests would be useful by itself.

PR #453, opened 2021-10-11
[GIT PULL] .github/workflows/build.yml: Add nolibc build x86-64 for CI

ammarfaizi2:
Hi Jens,

Please pull one commit to add x86-64 nolibc build for CI from me.
v2: Rebased based on your master branch.

I didn't think it was wrong history, but I have just pushed a new branch, this should match with the history.
```
The following changes since commit c7d03dc2c123bc2ee85582d8f922acd268d6ac78:

  configure: Add `CONFIG_NOLIBC` variable and macro (2021-10-10 09:09:56 -0600)

  https://github.com/ammarfaizi2/liburing add-ci-nolibc-x86-64

for you to fetch changes up to e4b1317c35684cd14cc21b5176ec39d56cbcc329:

  .github/workflows/build.yml: Add nolibc build x86-64 for CI (2021-10-11 08:41:58 +0700)

Ammar Faizi (1):
      .github/workflows/build.yml: Add nolibc build x86-64 for CI

 .github/workflows/build.yml | 10 ++++++----
 1 file changed, 6 insertions(+), 4 deletions(-)
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,10 +33,12 @@ jobs:
     - name: Build
       run: |
         ./configure --cc=${{matrix.cc}} --cxx=${{matrix.cxx}};
-        make V=1 -j$(nproc) \
-             CPPFLAGS="-Werror" \
-             CFLAGS="$FLAGS" \
-             CXXFLAGS="$FLAGS";
+        make -j$(nproc) V=1 CPPFLAGS="-Werror" CFLAGS="$FLAGS" CXXFLAGS="$FLAGS";
+
+    - name: Build nolibc x86-64
+      run: |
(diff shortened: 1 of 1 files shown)


ammarfaizi2:
> Note the io_uring_free_probe() commit, there's a bunch of private history in your tree. Hence it'll through a ton of merge failures when you attempt to pull it.

Ah right, it seems I have an identical name with tag that is not relevant with master branh history since the RFC.

ammarfaizi2:
It turned my branch was correct, but there is a tag that has the same name with my branch. This tag indeed has the wrong history. And what you showed is the tag `(tag: nolibc-x86-64)`. Lol.

Moving on...
Ignore it.

PR #456, opened 2021-10-11
setup.c: handmade fls -> __builtin_clz

Tindarid:
I think such change makes it more clear

### DIFF
--- a/src/setup.c
+++ b/src/setup.c
@@ -211,33 +211,11 @@ void io_uring_free_probe(struct io_uring_probe *probe)
 	free(probe);
 }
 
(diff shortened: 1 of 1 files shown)


axboe:
Why not just use __builtin_ffs() instead?

axboe:
OK I'm an idiot! Misremembered what it was doing. I'm fine with your change. But let's make it:

```
if (!x)
    return 0;
return 8 * sizeof(int) - __builtin_clz(x);
```

I hate reading ternarys.

Tindarid:
fixed

axboe:
Thanks!

Issue #461, opened 2021-10-20
forgotten flags argument for IORING_OP_WRITEV ?


romange:
Man page says that IORING_OP_WRITEV  reflects pwritev2(2). However, this call has `flags` argument that is omitted by the `io_uring_prep_writev`  function in liburing.h. Is it intentional?

axboe:
Hmmm yes, that should have been in the helper for that. You can just set sqe->rw_flags after calling the prep helper, that's where those flags go.

PR #462, opened 2021-10-21
Add io_uring_prep_readv2 and io_uring_prep_writev2 helpers 

romange:
these correspond to preadv2 and pwritev2 calls.
Completes #461


ammarfaizi2:
@romange Please use `Signed-off-by` tag.

The format for commit message is:
- First line is title (don't be too long like that if possible, put it in description for further detail).
- Empty line.
- Description (may be omitted for trivial changes).
- Empty line again (if it has description).
- Then a `Signed-off-by` with your name and email. For example:
```
```

Description should be word wrapped 72 chars. Some things should not be word-wrapped, they may be some kind of quoted text - long compiler error messages, oops reports, URL, or whatever things that have a certain specific format.

If you want to reference an issue or PR, use `Link` tag, in your case, it can be:
```
Link: https://github.com/axboe/liburing/issues/461
```
See this one for the example: https://git.kernel.dk/cgit/liburing/commit/?id=abd495348b5e0efee0a87b9206174a232932c21e

Issue #443, opened 2021-09-27
An option to use liburing without libc?


mrakh:
There are a couple of issues with `liburing`'s `libc` dependency:

- `libc` implementations of things like `errno` and `malloc` pollute the code with global/thread-local state. This makes reentrancy impossible and relocations expensive.
- `libc` doesn't play nice with non-POSIX threading models, like green threads with small stack sizes, or direct use of the `clone()` system call. This makes interop with non-C languages/runtimes difficult.

One could use the raw syscall interface to `io_uring` to address these concerns, but that would be somewhat painful, so it would be nice for `liburing` to support this use case out of the box. Perhaps something like a `NOLIBC` macro could be added which, if defined, would patch out `libc` constructs and replace them with non-`libc` wrappers where applicable. A few API changes might be necessary for the non-`libc` case (e.g. `io_uring_get_probe`/`io_uring_free_probe`), but it shouldn't break existing applications as long as it's opt-in.

YoSTEALTH:
As far as I know liburing has no dependency! That said I did notice liburing released by the OS distributors add `libc/glibc` as its dependency. You could just download liburing and compile it yourself.

ammarfaizi2:
Sounds good, I actually have the same problem but never really brought it up here.

> A few API changes might be necessary for the non-`libc` case (e.g. `io_uring_get_probe`/`io_uring_free_probe`), but it shouldn't break existing applications as long as it's opt-in.

Perhaps we can use VM directly with `mmap()` / `munmap()` syscall to replace `malloc()` / `free()` for the no libc case. That shouldn't be too hard to be done. I would be happy to volunteer to bring this feature to liburing.

ammarfaizi2:
> As far as I know liburing has no dependency! That said I did notice liburing released by the OS distributors add `libc/glibc` as its dependency. You could just download liburing and compile it yourself.

liburing *does* depend on libc, the use of `malloc()` / `free()`, `errno` are the examples of it.

ammarfaizi2:
> > As far as I know liburing has no dependency! That said I did notice liburing released by the OS distributors add `libc/glibc` as its dependency. You could just download liburing and compile it yourself.
> 
> liburing _does_ depend on libc, the use of `malloc()` / `free()`, `errno` are the examples of it.

Here is the shared library `ldd` output.
```
ammarfaizi2@integral:~/project/now/liburing/src$ ldd liburing.so.2.1.0 
[...]

PR #484, opened 2021-11-23
[WIP][GIT PULL] Always use `-fpic -fPIC` flag for nolibc build

ammarfaizi2:
Hi Jens,

This PR only contains 1 commit from me.

Bob reports linking time error for nolibc build:
```
  relocation R_X86_64_PC32 against symbol `memset' can not be \
  used when making a shared object; recompile with -fPIC
```
When we statically link liburing.a into a shared object, we need to
make sure it's built with `-fpic -fPIC` flag.

Note:
liburing.a can be used to create a new shared object (*.so) or a
non shared object ELF. When we use it to create a shared object,
it needs to be compiled with the above flag.

Always using this flag for nolibc build even if the target is not
a shared object ELF is still fine.

Link: https://github.com/axboe/liburing/issues/443#issuecomment-975493681
Fixes: c7d03dc2c123bc2ee85582d8f922acd268d6ac78 ("configure: Add `CONFIG_NOLIBC` variable and macro")
Fixes: f48ee3168cdc325233825603269f304d348d323c ("Add nolibc build support")
```
Reported-by: Bob Chen <beef9999@qq.com>
Tested-by: Bob Chen <beef9999@qq.com>
```
Cc: @beef9999

----
## git request-pull output:
```
The following changes since commit df9a741895028204b130b6cbe03c5c3dda3cf1a9:

  Rename man page (2021-11-18 07:27:17 -0700)

  https://github.com/ammarfaizi2/liburing tags/pic-issue-nolibc

for you to fetch changes up to bfb8975dda09f39d9f947f7b431f3dc01cb74b78:

  src/Makefile: Always use `-fpic -fPIC` flag for nolibc build (2021-11-23 16:23:40 +0700)

pic-issue-nolibc

Ammar Faizi (1):
      src/Makefile: Always use `-fpic -fPIC` flag for nolibc build

 src/Makefile | 6 +++---
 1 file changed, 3 insertions(+), 3 deletions(-)
```
----

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -36,9 +36,9 @@ liburing_srcs := setup.c queue.c register.c
 
 ifeq ($(CONFIG_NOLIBC),y)
 	liburing_srcs += nolibc.c
-	override CFLAGS += -nostdlib -nodefaultlibs -ffreestanding -fno-stack-protector
-	override CPPFLAGS += -nostdlib -nodefaultlibs -ffreestanding -fno-stack-protector
-	override LINK_FLAGS += -nostdlib -nodefaultlibs
+	override CFLAGS += -fpic -fPIC -nostdlib -nodefaultlibs -ffreestanding -fno-stack-protector
+	override CPPFLAGS += -fpic -fPIC -nostdlib -nodefaultlibs -ffreestanding -fno-stack-protector
+	override LINK_FLAGS += -fpic -fPIC -nostdlib -nodefaultlibs
 else
 	liburing_srcs += syscall.c
 endif


beef9999:
@ammarfaizi2 The new .a lib caused memory leak in my old program. I didn't even start to use this lib yet, just linked it.

Please hold merging this PR until I figured out what happened.

ammarfaizi2:
> @ammarfaizi2 The new .a lib caused memory leak in my old program. I didn't even start to use this lib yet, just linked it.
> 
> Please hold merging this PR until I figured out what happened.

Sure, please send any update with the progress.

beef9999:
@ammarfaizi2 I realized that the symbols from this liburing.a will replace the global glibc `malloc` `memset`, which might be the root cause of the mem leak. So I started to wonder there are better options than --nolibc.

After some dig, I found that gcc has a [`--version-script`](https://www.gnu.org/software/gnulib/manual/html_node/LD-Version-Scripts.html) parameter, which will export symbols for a minimum requirement based on liburing.map. No symbols conflicts any more.

So .a static lib is working now.

beef9999:
@ammarfaizi2  My opinion is `--version-script` could solve everything we discussed. The `nolibc` seems not required any more.

ammarfaizi2:
> The nolibc seems not required any more.

What do you mean by "The nolibc seems not required any more."?

`--version-script` can't replace `--nolibc`. `--nolibc` is arch dependent build that does not link `libc` shared object.

beef9999:
@ammarfaizi2 I mean usually people need an independent lib because they don't wish to meet symbol conflicts. Once this problem solved, the independency is not needed.

But perhaps due to my limit knowledge, I didn't come up with other scenarios that --nolibc is needed. Would you elaborate?

ammarfaizi2:
> @ammarfaizi2 I mean usually people need an independent lib because they don't wish to meet symbol conflicts. Once this problem solved, the independency is not needed.

You are correct. Apart from that, there is also another case, people may also not want to link libc to their project.

> But perhaps due to my limit knowledge, I didn't come up with other scenarios that --nolibc is needed. Would you elaborate?

The dependencies are not only `malloc()` and `memset()`, it includes every syscall used in that project. Like `mmap(2)`, `close(2)`, `munmap(2)`, `syscall(2)`, etc.

Apparently, those syscalls need libc wrapper to work with. What if we don't want libc to be linked in our project?

`--nolibc` comes as a solution to that question. If people don't want to use libc, they have to build with `--nolibc` flag, it will construct Assembly code directly to implement those syscalls. Unlike libc wrapper, Assembly code is not portable, so we have to code every single architecture. Currently supported arch to build nolibc is only x86-64.

ammarfaizi2:
See the discussion in this issue for detail https://github.com/axboe/liburing/issues/443#issue-1007905107

ammarfaizi2:
Let's close this, now I think always using `-fpic -fPIC` is not a proper solution. We should prevent `malloc()`, `memset()` replace libc function in case the target is compiled with libc. That's not what we do here.

Issue #490, opened 2021-12-04
Update operations truncate the provided user_data to 32bit on 32bit builds.


rbernon:
The internal structures use `__u64` both for `addr` field and for `user_data` fields, which is nice as they are usually matched against each other for update operations.

However, the internal `io_uring_prep_rw` function takes a `void * addr`, (but then casts it back to `__u64`), Several other public API functions, are doing truncation, such as `io_uring_prep_poll_update`, which takes `void* old_user_data` / `void* new_user_data`, or even worse, `io_uring_prep_timeout_update` which takes a `__u64 user_data` to match the timeout that needs updating but internally casts it to `unsigned long` / `void *` when calling `io_uring_prep_rw`.

Overall, this makes the `liburing` API fairly unusable on 32bit for such operations, as the `addr` field get truncated, and the user needs to fill the sqe fields by hand. This is pretty inconvenient for writing portable code, and the API should make sure no unnecessary truncation happens.

axboe:
What cases on 32-bit wants to put a 64-bit address into what should be a user pointer? Looking at this more closely, I'm a bit puzzled on what commands are involved here? Off the top of my head I don't recall any commands actually not using that for an actual user address, in which case how could that even work to put a non-valid 32-bit pointer in there?

Is this just a timeout command (and/or update) issue? Are there others? From a quick look, looks like timeout_update and timeout_remove are the only problematic ones here.

axboe:
And if so, why don't we just do:

```
diff --git a/src/include/liburing.h b/src/include/liburing.h
index 169e0983a276..230cd6195689 100644
--- a/src/include/liburing.h
+++ b/src/include/liburing.h
@@ -425,8 +425,8 @@ static inline void io_uring_prep_timeout(struct io_uring_sqe *sqe,
 static inline void io_uring_prep_timeout_remove(struct io_uring_sqe *sqe,
 						__u64 user_data, unsigned flags)
 {
-	io_uring_prep_rw(IORING_OP_TIMEOUT_REMOVE, sqe, -1,
-				(void *)(unsigned long)user_data, 0, 0);
+	io_uring_prep_rw(IORING_OP_TIMEOUT_REMOVE, sqe, -1, NULL, 0, 0);
+	sqe->addr = user_data;
 	sqe->timeout_flags = flags;
 }
 
@@ -434,9 +434,9 @@ static inline void io_uring_prep_timeout_update(struct io_uring_sqe *sqe,
 						struct __kernel_timespec *ts,
[...]

PR #491, opened 2021-12-04
liburing.h: Prevent unwanted addr 32bit truncations.

rbernon:
The addr field is used by several operations to match against a previous user_data field. Using void * for the io_uring_prep_rw parameter type causes undesired 32bit truncation when building the library for 32bit architecture and may fail to match a previously set 64bit user_data.

I think something like that should be done to fix https://github.com/axboe/liburing/issues/490

### DIFF
--- a/examples/io_uring-cp.c
+++ b/examples/io_uring-cp.c
@@ -74,7 +74,7 @@ static void queue_prepped(struct io_uring *ring, struct io_data *data)
 	else
 		io_uring_prep_writev(sqe, outfd, &data->iov, 1, data->offset);
 
-	io_uring_sqe_set_data(sqe, data);
+	io_uring_sqe_set_data(sqe, (__u64) (uintptr_t) data);
 }
 
 static int queue_read(struct io_uring *ring, off_t size, off_t offset)
@@ -100,7 +100,7 @@ static int queue_read(struct io_uring *ring, off_t size, off_t offset)
 	data->first_len = size;
 
 	io_uring_prep_readv(sqe, infd, &data->iov, 1, offset);
-	io_uring_sqe_set_data(sqe, data);
+	io_uring_sqe_set_data(sqe, (__u64) (uintptr_t) data);
 	return 0;
 }
 
@@ -185,7 +185,7 @@ static int copy_file(struct io_uring *ring, off_t insize)
 			if (!cqe)
 				break;
 
-			data = io_uring_cqe_get_data(cqe);
+			data = (void *) (uintptr_t) io_uring_cqe_get_data(cqe);
 			if (cqe->res < 0) {
 				if (cqe->res == -EAGAIN) {
 					queue_prepped(ring, data);
@@ -235,7 +235,7 @@ static int copy_file(struct io_uring *ring, off_t insize)
 			fprintf(stderr, "write res=%d\n", cqe->res);
 			return 1;
 		}
-		data = io_uring_cqe_get_data(cqe);
+		data = (void *) (uintptr_t) io_uring_cqe_get_data(cqe);
 		free(data);
 		writes--;
 		io_uring_cqe_seen(ring, cqe);
--- a/examples/link-cp.c
+++ b/examples/link-cp.c
@@ -79,16 +79,16 @@ static void queue_rw_pair(struct io_uring *ring, off_t size, off_t offset)
 	sqe = io_uring_get_sqe(ring);
 	io_uring_prep_readv(sqe, infd, &data->iov, 1, offset);
 	sqe->flags |= IOSQE_IO_LINK;
-	io_uring_sqe_set_data(sqe, data);
+	io_uring_sqe_set_data(sqe, (__u64) (uintptr_t) data);
 
 	sqe = io_uring_get_sqe(ring);
 	io_uring_prep_writev(sqe, outfd, &data->iov, 1, offset);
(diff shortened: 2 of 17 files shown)


axboe:
This is sure to trigger a bunch of warnings on compile now, for both 32-bit and 64-bit. Did you compile it?

Also please make sure to include a Signed-off-by line at the end of the commit, with your real name and email.

rbernon:
I did, but I think I have it compiled already and the tests don't have a dependency on the header, so they were not recompiled and I missed the warnings there.

axboe:
My main worry here is not so much the tests, as they are under liburing and we can just fix them up, it's more applications currently using liburing and now throwing potentially a bunch of warnings when the library is updated. Need to ponder what the best approach is here.

rbernon:
Well, I understand the worry but I don't really see any good way to fix it. A good number of functions are just entirely unusable with 32bit pointers, and even worse, they will silently accept the parameters but not operate as expected.

The only other sane solution would be to deprecate all the problematic entry points to make new ones, which I'm not sure is much better. As I suspect io_uring and liburing are still pretty young I'm not sure it's worth the trouble (although that's up to you).

I think the warnings are harmless in any case, as it's all going to be cast to an integer type equal or larger than the current pointer arguments, and at least people will know what to fix.

rbernon:
Updated with signoff, fixes for the tests and examples, as well as changing the types in `io_uring_cqe_set_data` / `io_uring_cqe_get_data` too for consistency/

axboe:
> as well as changing the types in io_uring_cqe_set_data / io_uring_cqe_get_data too for consistency.

Let's please not do that, that's just unnecessary breakage. For those, I'd say just add io_uring_cqe_set_data64 and ditto on the get side.

axboe:
The problem is really only if you want to use 64-bit types, if used for pointers then everything works just fine on 32-bit as-is. Should it work with 64-bit? Certainly! But it's not really a huge issue, at least nobody has complained about it yet :-)

> The only other sane solution would be to deprecate all the problematic entry points to make new ones, which I'm not sure is much better. As I suspect io_uring and liburing are still pretty young I'm not sure it's worth the trouble (although that's up to you).

It is quite new, but it's still got a few years on its back  by now. If used with -Werror or similar, then it will break compile. At least it's a trivial fix. I'd probably lean towards fixing them up and throwing the warnings, I do wonder if it should be accompanied with a define or similar so applications at least have a chance to compile without warnings on regardless of liburing version.

It's not that it's a hard fix, it's trivial. But it is introduced breakage, which I try to always void.

rbernon:
I'll probably leave it to you then, but in my opinion and with the current state of the `liburing` api, you should definitely break things, and probably better now rather than later when more people will be using it. It is implicitly broken already, and that needs to be made obvious so that anyone who may have missed that, gets informed.

For the same reason, I don't think you should keep the get_data / set_data helpers untouched: the sqe structure is accessible, and there's even a lot of code out there, in examples, snippets, etc, which are advertising the direct use of `user_data`, packing things in it because it's convenient and more efficient than using a pointer in a lot of user cases. These helpers will always be broken when used at the same time and that should be prevented.

Maybe another way of cleaning things up would be to change that `user_data` member, hiding it or retyping it to `void *` like all the rest of the code seems to assume, but you can't have a 64bit member here when operations that are referencing it only support a 32bit value. It is too slippery. Then on the other hand, like I said above, having a 64bit `user_data` member is convenient and pretty useful, and as the kernel supports it, it would be a shame not to expose it.

axboe:
I do agree that it should get fixed, no arguments on that one. I think where we disagree somewhat is the importance of it. I suspect the wine background means the 32-bit nature means a lot more to you, it's generally not something that anyone else has run into as most folks are on 64-bit anyway. And if they are on 32-bit, then storing pointers works just fine. It's only really a potential issue if you are running 32-bit and want to use the full 64-bit user_data space.

So while I do think that this should get fixed up, it's also potential pain for 99% of the users, that currently have no issues. Which is why I'm a bit hesitant to just do a flag day change like this.

user_data is part of the kernel-user ABI and cannot be changed, and you cannot have pointers in a shared struct like that or 32 vs 64-bit compatibility would be broken anyway. That's a potentially larger issue, with 32-bit applications on 64-bit kernels.

The changes might be more palatable if broken down somewhat, like first fixing up prep_rw. I'll take a stab at it.

rbernon:
> user_data is part of the kernel-user ABI and cannot be changed, and you cannot have pointers in a shared struct like that or 32 vs 64-bit compatibility would be broken anyway. That's a potentially larger issue, with 32-bit applications on 64-bit kernels.

Well the structure definition could be duplicated in `liburing`, exposing a different type for `user_data` for better compatibility with the way you like the library to be used, while keeping the alignment with an union or something else.

Regarding the 32bit concerns, sure, it's less of a concern these days and most people don't notice, but I'm sure `liburing` users include or will include libraries that are meant to be portable, and although it's becoming less likely to be used or even tested on 32bit architectures, they may still be supporting them.

Even Wine is trying to move away from running pure 32bit applications, but I don't think we're free of it yet, and not before a long time.

axboe:
Can you check current -git? I tried to do the bare minimum, and I tested it with both 32-bit and 64-bit applications on a 64-bit kernel.

rbernon:
Looks good, thanks.

PR #506, opened 2021-12-23
[GIT PULL] Clarify behaviour of offset in `prep_read`/`prep_write`.

ioquatix:
----
## git request-pull output:
```
The following changes since commit 7233a8753cb7a4a137cdc5e534bb31fb2a0d53fd:

  test/getdents: ignore run if argument specified (2021-12-22 19:54:45 -0700)

  git@github.com:ioquatix/liburing.git man-page-offset-clarification

for you to fetch changes up to 9ec56b9fb31d9f77fc23113ceda27d7b4cbad19e:

  Clarify behaviour of offset in `prep_read`/`prep_write`. (2021-12-23 17:29:24 +1300)

Samuel Williams (1):
      Clarify behaviour of offset in `prep_read`/`prep_write`.

 man/io_uring_prep_read.3    | 17 +++++++++++++++--
 man/io_uring_prep_readv.3   | 16 +++++++++++++++-
 man/io_uring_prep_readv2.3  | 19 +++++++++++++++++--
 man/io_uring_prep_write.3   | 17 +++++++++++++++--
 man/io_uring_prep_writev.3  | 18 ++++++++++++++++--
 man/io_uring_prep_writev2.3 | 21 +++++++++++++++++----
 6 files changed, 95 insertions(+), 13 deletions(-)
```
----
<details>

## Pull Request Guidelines
```
```

2. Then an empty line.
```
```

certain specific format).

of truth, not github.

URL.

```
```

Use a real email address!

### Commit message example:
```

safe without synchronization.

```

</details>

----

### DIFF
--- a/man/io_uring_prep_read.3
+++ b/man/io_uring_prep_read.3
@@ -26,8 +26,16 @@ to start reading
 .I nbytes
 into the buffer
 .I buf
-at the
+at the specified
 .I offset.
+
+On files that support seeking, if the offset is set to -1, the read operation
+commences at the file offset, and the file offset is incremented by the number
+of bytes read. See
+.BR read (2)
+for more details.
+
+On files that are not capable of seeking, the offset is ignored.
 
 After the read has been prepared it can be submitted with one of the submit
 functions.
--- a/man/io_uring_prep_readv.3
+++ b/man/io_uring_prep_readv.3
@@ -26,11 +26,20 @@ to start reading
 .I nr_vecs
 into the
 .I iovecs
-array at
+array at the specified
 .I offset.
+
+On files that support seeking, if the offset is set to -1, the read operation
+commences at the file offset, and the file offset is incremented by the number
+of bytes read. See
+.BR read (2)
+for more details.
+
+On files that are not capable of seeking, the offset is ignored.
 
 After the write has been prepared it can be submitted with one of the submit
 functions.
+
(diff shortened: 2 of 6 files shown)


ioquatix:
@axboe I thought we could clarify the offset argument. Let me know what you think.

axboe:
I'm fine with clarifying it, but file descriptor is well understood and means any file type. Changing it to object makes it a lot less understandable imho.

axboe:
The non file descriptor changes look great, would you mind doing just those?

ioquatix:
> I'm fine with clarifying it, but file descriptor is well understood and means any file type. Changing it to object makes it a lot less understandable imho.

I was following the BSD man pages which use the term object but I see the Linux man pages are different:

https://www.freebsd.org/cgi/man.cgi?sektion=2&query=read

Not sure which one makes more sense.

ioquatix:
I think calling everything a "file" is confusing. I think this is probably why BSD uses the term object, since there are a lots of things that can be referenced by a "file descriptor". In my own code, I generally don't refer to it as "fd" or "file descriptor" but now just refer to it as "descriptor" to try and avoid this cognitive overload.

ioquatix:
I rewrote it following Linux conventions which makes more sense.

axboe:
Looks good, thanks!

Issue #516, opened 2022-01-28
statx configure and build issues with older toolchain


JonKohler:
Trying to integrate liburing into an older-ish centos7 based toolchain, for a product that we have that is not quite yet ready for a larger toolchain update.

On latest master, when I do this configure, then make, it fails in two separate ways: `./configure --cxx=/opt/cross/el7.6-x86_64/gcc-7.4.0/bin/g++ --cc=/opt/cross/el7.6-x86_64/gcc-7.4.0/bin/gcc --nolibc` 

First, this will not configure, with following error: 
```
sh-4.2# ./configure --cxx=/opt/cross/el7.6-x86_64/gcc-7.4.0/bin/g++ --cc=/opt/cross/el7.6-x86_64/gcc-7.4.0/bin/gcc --nolibc
prefix                        /usr
includedir                    /usr/include
libdir                        /usr/lib
libdevdir                     /usr/lib
relativelibdir                
mandir                        /usr/man
datadir                       /usr/share
stringop_overflow             yes
array_bounds                  yes
__kernel_rwf_t                no
__kernel_timespec             no
open_how                      no
statx                         no
glibc_statx                   no
C++                           yes
has_ucontext                  yes
has_memfd_create              no
liburing_nolibc               yes
CC                            /opt/cross/el7.6-x86_64/gcc-7.4.0/bin/gcc
CXX                           /opt/cross/el7.6-x86_64/gcc-7.4.0/bin/g++
./configure: line 474: no: command not found
```

This appears to be the way that the test $glibc_statx line was written in configure. When I fix that to read: `if [ "$glibc_statx" = "no" ] && [ "$statx" = "yes" ]; then`

It then evaluates/configures properly; however, when I then run make, I get this:

```
sh-4.2# make
...
     CC skip-cqe
statx.c: In function 'test_statx':
statx.c:43:15: error: storage size of 'x1' isn't known
  struct statx x1, x2;
               ^~
statx.c:43:19: error: storage size of 'x2' isn't known
  struct statx x1, x2;
                   ^~
statx.c:51:40: error: 'STATX_ALL' undeclared (first use in this function); did you mean 'STA_PLL'?
  io_uring_prep_statx(sqe, -1, path, 0, STATX_ALL, &x1);
                                        ^~~~~~~~~
                                        STA_PLL
statx.c:51:40: note: each undeclared identifier is reported only once for each function it appears in
statx.c:43:19: warning: unused variable 'x2' [-Wunused-variable]
[...]

PR #517, opened 2022-01-28
[GIT PULL] configure: fix statx configure and build on !CONFIG_HAVE_STATX toolchains

JonKohler:
Update configure and test/Makefile, such that if !CONFIG_HAVE_STATX will
not generate statx.c test. Without this patch, liburing fails to compile
on an older toolchain that does not have statx available.

Fixes: #516

Signed-off-by: Jon Kohler <jon@nutanix.com>

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -138,7 +138,6 @@ test_srcs := \
 	sq-poll-share.c \
 	sqpoll-sleep.c \
 	sq-space_left.c \
-	statx.c \
 	stdout.c \
 	submit-link-fail.c \
 	submit-reuse.c \
--- a/configure
+++ b/configure
@@ -471,7 +471,7 @@ else cat >> $compat_h << EOF
 
 EOF
 fi
-if test "$glibc_statx" = "no" && "$statx" = "yes"; then
+if [ "$glibc_statx" = "no" ] && [ "$statx" = "yes" ]; then
 cat >> $compat_h << EOF
 #include <sys/stat.h>
 


test/Makefile
```diff
@@ -138,7 +138,6 @@ test_srcs := \
 	sq-poll-share.c \
 	sqpoll-sleep.c \
 	sq-space_left.c \
-	statx.c \
```
axboe, line 141: This doesn't look right?

JonKohler: statx.c gets added if CONFIG_HAVE_STATX on line 161

axboe:
Don't think this fix is right. The statx tests generally should just figure out of glibc has the definition, and if it doesn't, we generate the compat bits for it. That means that any statx related test should just work, and the problem is more likely with when we generate the definition in compat.h or not.

test/Makefile
axboe: If the system doesn't have statx support, liburing should make its own definition. Maybe that got broken, in which case, that's what needs fixing. The statx test should always compile just fine.

JonKohler:
ok thanks for the quick feedback. At minimum though, the `configure` side of this change seems to make sense. if the system doesn't have statx, the `test` command does not evaluate correctly, right?

test/Makefile
axboe: Ah ok, I'm mistaken, we do have a check for it and add it conditionally. I guess it all looks fine then, thanks.

JonKohler:
Thanks @axboe - fastest merge request I've ever gotten in my entire life, good stuff!

axboe:
We aim to make everything around liburing/io_uring fast ;-)

PR #544, opened 2022-03-09
make: let src/Makefile set *dir vars properly

257:
prefix suffice for setting *dir variables in src/Makefile; this does
away with setting does in in ./Makefile
----
## git request-pull output:
```
The following changes since commit 6231f56da7881bde6fb011e1b54d672f8fe5a224:

  src/queue: don't keep looping for IOPOLL peek operation (2022-03-05 15:47:26 -0700)

  https://github.com/257/liburing.git makefile

for you to fetch changes up to c1f157802c72e8a331710e675e843e6aa4c2c7ad:

  [GIT PULL] make: let src/Makefile set *dir vars properly (2022-03-09 15:51:31 -0500)

Paymon MARANDI (1):
      [GIT PULL] make: let src/Makefile set *dir vars properly

 Makefile | 6 +-----
 1 file changed, 1 insertion(+), 5 deletions(-)
```
----
<details>

## Pull Request Guidelines
```
```

2. Then an empty line.
```
```

certain specific format).

of truth, not github.

URL.

```
```

Use a real email address!

### Commit message example:
```

safe without synchronization.

```

</details>

----

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,9 +45,6 @@ endif
 
 install: $(NAME).pc
 	@$(MAKE) -C src install prefix=$(DESTDIR)$(prefix) \
-		includedir=$(DESTDIR)$(includedir) \
-		libdir=$(DESTDIR)$(libdir) \
-		libdevdir=$(DESTDIR)$(libdevdir) \
 		relativelibdir=$(relativelibdir)
 	$(INSTALL) -D -m 644 $(NAME).pc $(DESTDIR)$(libdevdir)/pkgconfig/$(NAME).pc
 	$(INSTALL) -m 755 -d $(DESTDIR)$(mandir)/man2


Makefile
```diff
@@ -44,11 +44,7 @@ endif
 	    $< >$@
 
 install: $(NAME).pc
-	@$(MAKE) -C src install prefix=$(DESTDIR)$(prefix) \
-		includedir=$(DESTDIR)$(includedir) \
-		libdir=$(DESTDIR)$(libdir) \
-		libdevdir=$(DESTDIR)$(libdevdir) \
-		relativelibdir=$(relativelibdir)
+	@$(MAKE) -C src install prefix=$(DESTDIR)$(prefix)
```
ammarfaizi2, line 47: The prefix only applies to these 3:
```
includedir ?= $(prefix)/include
libdir ?= $(prefix)/lib
libdevdir ?= $(prefix)/lib
```

You shouldn't get rid of the `relativelibdir` here. What do you think?

ammarfaizi2 (changes requested):
The `[GIT PULL]` prefix is for the pull request title, not in the commit message. Please reword that.

Makefile
257: > You shouldn't get rid of the `relativelibdir` here. What do you think?

that gets deduces from `$libdir` in `./configure`, it seems, although it's only used when `ENABLE_SHARED` in switched on and i don't see that check in `./configure`, maybe @stefano-garzarella can shed some light on this?

leaving it there for now.

257:
done!

PR #547, opened 2022-03-14
[GIT PULL] liburing.h: define GNU_SOURCE for cpu_set_t

thesamesam:
On musl, cpu_set_t is only exposed if GNU_SOURCE is defined. While in
the liburing build system, this is set (43b7ec8d17888df0debccda27dd58f4d1b90245e),
it can't be guaranteed that folks including the header externally will set
that macro.

Noticed while investigating a build failure for glusterfs on a musl
system:
```
configure:17701: checking for liburing.h
configure:17701: x86_64-gentoo-linux-musl-gcc -c -pipe -march=native -fno-diagnostics-color -O2  conftest.c >&5
In file included from conftest.c:105:
/usr/include/liburing.h:162:39: error: unknown type name 'cpu_set_t'
  162 |                                 const cpu_set_t *mask);
      |                                       ^~~~~~~~~
configure:17701: $? = 1
```

Just like _XOPEN_SOURCE, set if needed.

Bug: https://bugs.gentoo.org/829293
See: 43b7ec8d17888df0debccda27dd58f4d1b90245e
Signed-off-by: Sam James <sam@gentoo.org>

----
## git request-pull output:
```
The following changes since commit d88bc35e0b971f78ca1f4478cfa2c0833c88740e:

  Add io_uring_register_iowq_max_workers() man page (2022-03-13 17:05:53 -0600)

  https://github.com/thesamesam/liburing musl-gnu-source

for you to fetch changes up to 131086fa30de0a544a19d6018d11c77cf00cb40b:

  liburing.h: define GNU_SOURCE for cpu_set_t (2022-03-14 14:20:43 +0000)

Sam James (1):
      liburing.h: define GNU_SOURCE for cpu_set_t

 src/include/liburing.h | 4 ++++
 1 file changed, 4 insertions(+)
```
----
<details>

## Pull Request Guidelines
```
```

2. Then an empty line.
```
```

certain specific format).

of truth, not github.

URL.

```
```

Use a real email address!

### Commit message example:
```

safe without synchronization.

```

</details>

----


thesamesam:
Thank you both!

PR #553, opened 2022-04-01
[GIT PULL] Allow configure script to fail if there are bash errors on the way.

romange:
Signed-off-by: Roman Gershman  <romange@gmail.com>

For poorly configured environments, `configure` may fail but will still exit with code 0.
This, in turn, causes cryptic build errors later (happened to me on alpine linux without coreutils package installed).

----
## git request-pull output:
```
The following changes since commit 3b52ae980c41b420592e3005199041890a56d6dc:

  test/fixed-reuse: test that linked file slot reuse is correct (2022-03-30 11:33:38 -0600)

  https://github.com/romange/liburing.git 

for you to fetch changes up to 994f8d39ba42ea3171cb3bbee55ea08e6dda904b:

  configure script should fail if any of the underlying commands fail. (2022-04-01 15:04:52 +0300)

Roman Gershman (1):
      configure script should fail if any of the underlying commands fail.

 configure | 2 ++
 1 file changed, 2 insertions(+)

```
----
<details>

## Pull Request Guidelines
```
```

2. Then an empty line.
```
```

certain specific format).

of truth, not github.

URL.

```
```

Use a real email address!

### Commit message example:
```

safe without synchronization.

```

</details>

----

### DIFF
--- a/configure
+++ b/configure
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+set -e
+
 cc=${CC:-gcc}
 cxx=${CXX:-g++}
 


ammarfaizi2 (changes requested):
> For poorly configured environments, `configure` may fail but will still exit with code 0. This, in turn, causes cryptic build errors later (happened to me on alpine linux without coreutils package installed).

This explanation should also go into the commit message.

Another problem reported by the GitHub bot, this patch breaks the nolibc build, not sure what happens.

romange:
Maybe it has already been broken but "set -e" exposed it?

ammarfaizi2:
> Maybe it has already been broken but "set -e" exposed it?

Yeah, makes sense to me, it would be great if you can fix that as well instead of leaving the GitHub bot CI red.

romange:
Hi Ammar, unfortunately i wont be able to fix it, because i do not have
access to that type of environment at hand. I can close the PR if you
prefer.

ammarfaizi2:
> Hi Ammar, unfortunately i wont be able to fix it, because i do not have access to that type of environment at hand. I can close the PR if you prefer.

I will try to fix it and cherry-pick your commit. So yeah, let's close this. I will mention you in my PR later if I can figure it out.

ammarfaizi2:
> > Hi Ammar, unfortunately i wont be able to fix it, because i do not have access to that type of environment at hand. I can close the PR if you prefer.
> 
> I will try to fix it and cherry-pick your commit. So yeah, let's close this. I will mention you in my PR later if I can figure it out.

OK, I found the culprit.

In the configure script, we have:
```
optarg=$(expr "x$opt" : 'x[^=]*=\(.*\)')
```
If the $opt doesn't match with that regex, it will give exit code 1. So it will break any argument that doesn't have value.

Like `--help` and `--nolibc` for example.

I will send a PR to fix it soon.

PR #588, opened 2022-05-31
io_uring: add prep functions for current working directory

uarif1:
This is done for unlink, rename, mkdir, symlink and link.

Signed-off-by: Usama Arif <usama.arif@bytedance.com>

### DIFF
--- a/src/include/liburing.h
+++ b/src/include/liburing.h
@@ -18,6 +18,7 @@
 #include <stdbool.h>
 #include <inttypes.h>
 #include <time.h>
+#include <fcntl.h>
 #include <sched.h>
 #include <linux/swab.h>
 #include "liburing/compat.h"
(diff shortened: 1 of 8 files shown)


axboe:
Thanks for doing this. Two minor comments:

- Shouldn't the test/rename.c change itself actually use io_uring_prep_rename()?
- Can you do the man page additions as well? They should just go into the existing *at() man pages, no need for a separate page for them

src/include/liburing.h
```diff
@@ -739,6 +747,15 @@ static inline void io_uring_prep_renameat(struct io_uring_sqe *sqe, int olddfd,
 	sqe->rename_flags = (__u32) flags;
 }
 
+static inline void io_uring_prep_rename(struct io_uring_sqe *sqe,
```
axboe, line 750: I also think it'd be cleaner if this one just called io_uring_prep_renameat(), and ditto for the other ones. Then we keep the RENAMEAT sqe fields consistent by using the same parent helper.

YoSTEALTH:
wouldn't it be better to call `io_uring_prep_rw` directly? this way it actually makes it worth using `io_uring_prep_mkdir`


```c
static inline void io_uring_prep_mkdir(struct io_uring_sqe *sqe,
					const char *path, mode_t mode)
{
    io_uring_prep_rw(IORING_OP_MKDIRAT, sqe, AT_FDCWD, path, mode, 0);
}
```

vs

```c
static inline void io_uring_prep_mkdir(struct io_uring_sqe *sqe,
					const char *path, mode_t mode)
{
	io_uring_prep_mkdirat(sqe, AT_FDCWD, path, mode);
}
```

ammarfaizi2:
> wouldn't it be better to call `io_uring_prep_rw` directly? this way it actually makes it worth using `io_uring_prep_mkdir`

How come it be better? Those are the same thing. They are all inline anyway.

YoSTEALTH:
> How come it be better? Those are the same thing. They are all inline anyway.

Right, I overlooked it being `inline`.

PR #621, opened 2022-06-27
Exitcode protocol for reporting tests as skipped

eli-schwartz:
Inspired by the attempts at using other build systems with liburing, I took a look at how the testsuite is set up. I noticed that when tests are skipped due to lack of requirements in setting up the tests, they are reported as "passing" by the test harness, and the only real visibility into the fact that they were skipped is the log messages they emit.

GNU automake has a standard exitcode for tests like that -- just exit 77 -- and Meson respects it by default. (Even cmake can be made to respect it via `SKIP_RETURN_CODE`.) So it seems like a good idea to follow in the same footsteps as these existing practices. It's as good a choice as anything else, and has the benefit of consistency -- if the project ever switches build systems, this will already work without any further effort.

This PR changes runtests.sh to detect the special exitcode and list those tests together with "Tests skipped", and ports a few tests over to use it as well -- I don't thoroughly understand the test flow of many of the test programs, so I stuck to the obviously correct cases because a) it has to start somewhere, b) it demonstrates the basic idea of how to structure things.

### DIFF
--- a/test/runtests.sh
+++ b/test/runtests.sh
@@ -112,6 +112,9 @@ run_test()
 	if [ "$status" -eq 124 ]; then
 		echo "Test $test_name timed out (may not be a failure)"
 		TIMED_OUT="$TIMED_OUT <$test_string>"
+	elif [ "$status" -eq 77 ]; then
+		echo "Test $test_name skipped itself"
+		SKIPPED="$SKIPPED <$test_string>"
 	elif [ "$status" -ne 0 ]; then
 		echo "Test $test_name failed with ret $status"
 		FAILED="$FAILED <$test_string>"
@@ -164,6 +167,8 @@ fi
 if [ "${RET}" -ne 0 ]; then
 	echo "Tests failed: $FAILED"
 	exit $RET
+elif [ -n "$SKIPPED" ] && [ -n "$TEST_GNU_EXITCODE" ]; then
+	exit 77
 else
 	echo "All tests passed"
 	exit 0
--- a/test/ce593a6c480a.c
+++ b/test/ce593a6c480a.c
@@ -70,7 +70,7 @@ int main(int argc, char *argv[])
 	/* Setup the ring with a registered event fd to be notified on events */
 	ret = t_create_ring_params(8, &ring, &p);
 	if (ret == T_SETUP_SKIP)
-		return 0;
+		return 77;
 	else if (ret < 0)
 		return ret;
 
--- a/test/d4ae271dfaae.c
+++ b/test/d4ae271dfaae.c
@@ -31,7 +31,7 @@ int main(int argc, char *argv[])
 	p.flags = IORING_SETUP_SQPOLL;
 	ret = t_create_ring_params(4, &ring, &p);
 	if (ret == T_SETUP_SKIP)
-		return 0;
+		return 77;
 	else if (ret < 0)
 		return 1;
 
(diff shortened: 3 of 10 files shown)


axboe:
We already have some vague notion of SKIP for the helpers that setup eg the ring, maybe just build on that and make that the first class citizen? An enum always preferable than a define for this, as compilers can catch missing checks.

eli-schwartz:
Oh hmm, so if a special file/device is provided then the idea is to just run the subset of tests that actually test a specific device? I see.

Would it make sense to log that, e.g. `fprintf(stdout, "device provided but not a device-specific test, skipping\n")`?

axboe:
I don't think it should print any extra noise, that'll just be things I have to weed through when checking tests. They should just skip if passed an argument and they don't take one. Basically exactly what they do now, except we'll now distinguish between 'pass' and 'skipped'.

eli-schwartz:
Oops, added.

axboe:
Thanks!

axboe:
@eli-schwartz Are you going to cleanup the rest of the tests to ensure that we always pass back T_EXIT_*? We're kind of left in an indeterminate state right now, it's a bit of a mess to be honest. Needs to be pulled to completion.

eli-schwartz:
Yes. I can and will continue. As I stated in the original commits / Github cover letter, I am not always 100% sure how some exit codes are propagated from other functions -- it is however quite easy to review literal ints returned in `main()`, so I did some work on that, and did more work along the way while answering other review comments.

I wasn't going to insist you merge my patches with only a partial cleanup, but I'm not unhappy you did :) and I think it's reasonable to do this in an incremental manner. As I said on the mailing list -- I'll submit more work on this, but at the same time, new code can already do it properly, and other people can work on it too.

My plan is to work in alphanumeric order of test filenames, so if anyone else wants to work on it too, starting from "z" will optimize the chances that we don't contribute overlapping changes.

FTR -- I will be offline over the Sabbath, but my productivity usually spikes again on Sunday to make up for it.

test/runtests.sh
```diff
@@ -164,6 +167,8 @@ fi
 if [ "${RET}" -ne 0 ]; then
 	echo "Tests failed: $FAILED"
 	exit $RET
+elif [ -n "$SKIPPED" ] && [ -n "$TEST_GNU_EXITCODE" ]; then
+	exit 77
```
ddiss, line 171: Sorry to dig up an old (merged) PR, but I just came across this hunk and don't really know what it's there for - `TEST_GNU_EXITCODE` doesn't appear to have anything to do with automake tests. @eli-schwartz did you have a specific use for `TEST_GNU_EXITCODE`?

eli-schwartz: IIRC I used that when executing runtests.sh from meson instead of Make, and therefore had a test harness that understood 77.

The idea is that I could then set `TEST_GNU_EXITCODE=1` as an environment for runtests.sh, and then my harness would collect skipped tests and report them as skipped, but the Makefile simply runs it and exits with the exit code of runtests.sh itself and therefore wants skipped tests to exit 0.

ddiss: Ah okay, makes sense I suppose. Just keep in mind that this will also fire for tests which are skipped via `TEST_EXCLUDE` in `config.local`. Thanks for the quick response

eli-schwartz: That should be fine since skipped tests are considered passing, basically.

PR #645, opened 2022-08-28
[GIT PULL] queue: avoid io_uring_enter() with no SQEs

calebsander:
`io_uring_submit()` makes an `io_uring_enter()` syscall even if no SQEs are
ready to submit. Move the empty SQ check from `_io_uring_get_cqe()` to
`sq_ring_needs_enter()` so `io_uring_submit()` can skip the syscall too.

----
## git request-pull output:
```
The following changes since commit b1486931117f643b0ad6612beb3f9563b508126c:

  Merge branch 'refactor/cache-ring-size' of https://github.com/calebsander/liburing (2022-08-28 14:06:55 -0600)

  git@github.com:calebsander/liburing.git feature/no-empty-submit

for you to fetch changes up to 504efe18c213dd753fe6c69afbfff6cfd50ba4cc:

  queue: avoid io_uring_enter() with no SQEs (2022-08-28 19:04:53 -0600)

Caleb Sander (1):
      queue: avoid io_uring_enter() with no SQEs

 src/queue.c | 12 +++++++++---
 1 file changed, 9 insertions(+), 3 deletions(-)
```
----

### DIFF
--- a/src/queue.c
+++ b/src/queue.c
@@ -12,9 +12,15 @@
  * Returns true if we're not using SQ thread (thus nobody submits but us)
  * or if IORING_SQ_NEED_WAKEUP is set, so submit thread must be explicitly
  * awakened. For the latter case, we set the thread wakeup flag.
+ * If no SQEs are ready for submission, returns false.
  */
-static inline bool sq_ring_needs_enter(struct io_uring *ring, unsigned *flags)
+static inline bool sq_ring_needs_enter(struct io_uring *ring,
+				       unsigned submit,
+				       unsigned *flags)
 {
+	if (!submit)
+		return false;
+
 	if (!(ring->flags & IORING_SETUP_SQPOLL))
 		return true;
 
@@ -85,7 +91,7 @@ static int _io_uring_get_cqe(struct io_uring *ring,
 			flags = IORING_ENTER_GETEVENTS | data->get_flags;
 			need_enter = true;
 		}
-		if (data->submit && sq_ring_needs_enter(ring, &flags))
+		if (sq_ring_needs_enter(ring, data->submit, &flags))
 			need_enter = true;
 		if (!need_enter)
 			break;
@@ -354,7 +360,7 @@ static int __io_uring_submit(struct io_uring *ring, unsigned submitted,
 	int ret;
 
 	flags = 0;
-	if (sq_ring_needs_enter(ring, &flags) || wait_nr) {
+	if (sq_ring_needs_enter(ring, submitted, &flags) || wait_nr) {
 		if (wait_nr || (ring->flags & IORING_SETUP_IOPOLL))
 			flags |= IORING_ENTER_GETEVENTS;
 		if (ring->int_flags & INT_FLAG_REG_RING)


axboe:
This probably should be two changes - one for the syscall avoidance if there's nothing to submit, and one for the unsigned change. Would you mind splitting it into two? Whenever a commit message has an "Also ..." in there, that's usually a good sign that it should not be there but rather in a separate commit :)

axboe:
This one breaks IOPOLL, as we rely on entering the kernel for that always.

axboe:
Also see:

```
commit 1bafb3ce5f5eeb11cd982c7540f6aa74e3f381d4
Author: Jens Axboe <axboe@kernel.dk>
Date:   Thu Aug 20 21:40:16 2020 -0600

    sq_ring_needs_enter: revert change to only enter if submit != 0
    
    This reverts commit f0c5c54945ae92a00cdbb43bdf3abaeab6bd3a23.
```

which is basically identical to your change...

axboe:
I suspect we can get by with just checking for IOPOLL. Maybe...

axboe:
@calebsander ^^

just in case it gets lost in the noise and since it's closed.

calebsander:
Good point. I see we have `cq_ring_needs_enter()`, which is already used by `_io_uring_get_cqe()`:
```c
static inline bool cq_ring_needs_flush(struct io_uring *ring)
{
	return IO_URING_READ_ONCE(*ring->sq.kflags) &
				 (IORING_SQ_CQ_OVERFLOW | IORING_SQ_TASKRUN);
}

static inline bool cq_ring_needs_enter(struct io_uring *ring)
{
	return (ring->flags & IORING_SETUP_IOPOLL) || cq_ring_needs_flush(ring);
}
```
```c
if (!cqe && !data->wait_nr && !data->submit) {
	/*
	 * If we already looped once, we already entererd
	 * the kernel. Since there's nothing to submit or
	 * wait for, don't keep retrying.
	 */
	if (looped || !cq_ring_needs_enter(ring)) {
		err = -EAGAIN;
		break;
	}
	need_enter = true;
}
if (data->wait_nr > nr_available || need_enter) {
	flags = IORING_ENTER_GETEVENTS | data->get_flags;
	need_enter = true;
}
if (sq_ring_needs_enter(ring, data->submit, &flags))
	need_enter = true;
if (!need_enter)
	break;
```
I think this is probably the right thing to use instead of just checking the `IOPOLL` flag (since it also handles CQ overflow).

And I would move the logic to `__io_uring_submit()` instead of `sq_ring_needs_enter()`, since it doesn't really have to do with the SQ. Something like this?
```diff
diff --git a/src/queue.c b/src/queue.c
index 5eeaa0c..b52d88a 100644
--- a/src/queue.c
+++ b/src/queue.c
@@ -356,12 +356,13 @@ int io_uring_wait_cqe_timeout(struct io_uring *ring,
 static int __io_uring_submit(struct io_uring *ring, unsigned submitted,
                             unsigned wait_nr)
 {
+       bool cq_needs_enter = wait_nr || cq_ring_needs_enter(ring);
        unsigned flags;
        int ret;
 
        flags = 0;
-       if (sq_ring_needs_enter(ring, submitted, &flags) || wait_nr) {
-               if (wait_nr || (ring->flags & IORING_SETUP_IOPOLL))
+       if (sq_ring_needs_enter(ring, submitted, &flags) || cq_needs_enter) {
+               if (cq_needs_enter)
                        flags |= IORING_ENTER_GETEVENTS;
                if (ring->int_flags & INT_FLAG_REG_RING)
                        flags |= IORING_ENTER_REGISTERED_RING;
```

axboe:
Yep that looks more reasonable!

PR #646, opened 2022-08-28
[GIT PULL] man: clarify "never uses errno" refers to ring ops

calebsander:
"io_uring never uses `errno`" is too strong; io_uring syscalls set `errno`.
Clarify in the `io_uring_prep_*()` man pages that `errno` is not used
only for the results of ring operations.

----
## git request-pull output:
```
The following changes since commit b1486931117f643b0ad6612beb3f9563b508126c:

  Merge branch 'refactor/cache-ring-size' of https://github.com/calebsander/liburing (2022-08-28 14:06:55 -0600)

  git@github.com:calebsander/liburing.git fix/man-errno

for you to fetch changes up to 8769cd893a85bdb518ae99b7b9ddc7d55c10eaf3:

  man: clarify "never uses errno" refers to ring ops (2022-08-28 17:07:47 -0600)

Caleb Sander (1):
      man: clarify "never uses errno" refers to ring ops

 man/io_uring_prep_accept.3          | 4 ++--
 man/io_uring_prep_close.3           | 4 ++--
 man/io_uring_prep_connect.3         | 4 ++--
 man/io_uring_prep_fadvise.3         | 4 ++--
 man/io_uring_prep_fallocate.3       | 4 ++--
 man/io_uring_prep_fsync.3           | 4 ++--
 man/io_uring_prep_linkat.3          | 4 ++--
 man/io_uring_prep_madvise.3         | 4 ++--
 man/io_uring_prep_mkdirat.3         | 4 ++--
 man/io_uring_prep_openat.3          | 4 ++--
 man/io_uring_prep_openat2.3         | 4 ++--
 man/io_uring_prep_poll_add.3        | 4 ++--
 man/io_uring_prep_read.3            | 4 ++--
 man/io_uring_prep_read_fixed.3      | 4 ++--
 man/io_uring_prep_readv.3           | 4 ++--
 man/io_uring_prep_readv2.3          | 4 ++--
 man/io_uring_prep_recv.3            | 4 ++--
 man/io_uring_prep_recvmsg.3         | 4 ++--
 man/io_uring_prep_renameat.3        | 4 ++--
 man/io_uring_prep_send.3            | 4 ++--
 man/io_uring_prep_sendmsg.3         | 4 ++--
 man/io_uring_prep_shutdown.3        | 4 ++--
 man/io_uring_prep_socket.3          | 4 ++--
 man/io_uring_prep_splice.3          | 4 ++--
 man/io_uring_prep_statx.3           | 4 ++--
 man/io_uring_prep_symlinkat.3       | 4 ++--
 man/io_uring_prep_sync_file_range.3 | 4 ++--
 man/io_uring_prep_tee.3             | 4 ++--
 man/io_uring_prep_unlinkat.3        | 4 ++--
 man/io_uring_prep_write.3           | 4 ++--
 man/io_uring_prep_write_fixed.3     | 4 ++--
 man/io_uring_prep_writev.3          | 4 ++--
 man/io_uring_prep_writev2.3         | 4 ++--
 33 files changed, 66 insertions(+), 66 deletions(-)
```
----

### DIFF
--- a/man/io_uring_prep_accept.3
+++ b/man/io_uring_prep_accept.3
@@ -136,8 +136,8 @@ non-direct accept. Note that where synchronous system calls will return
 on failure and set
 .I errno
 to the actual error value, io_uring never uses
-.IR errno .
-Instead it returns the negated
+.IR errno
+for operations submitted to the ring. Instead it returns the negated
 .I errno
 directly in the CQE
 .I res
--- a/man/io_uring_prep_close.3
+++ b/man/io_uring_prep_close.3
@@ -47,8 +47,8 @@ details on possible values. Note that where synchronous system calls will return
 on failure and set
 .I errno
 to the actual error value, io_uring never uses
-.IR errno .
-Instead it returns the negated
+.IR errno
+for operations submitted to the ring. Instead it returns the negated
 .I errno
 directly in the CQE
 .I res
--- a/man/io_uring_prep_connect.3
+++ b/man/io_uring_prep_connect.3
@@ -44,8 +44,8 @@ details on possible values. Note that where synchronous system calls will return
 on failure and set
 .I errno
 to the actual error value, io_uring never uses
-.IR errno .
-Instead it returns the negated
+.IR errno
+for operations submitted to the ring. Instead it returns the negated
 .I errno
 directly in the CQE
 .I res
--- a/man/io_uring_prep_fadvise.3
+++ b/man/io_uring_prep_fadvise.3
@@ -46,8 +46,8 @@ details on possible values. Note that where synchronous system calls will return
 on failure and set
 .I errno
 to the actual error value, io_uring never uses
-.IR errno .
-Instead it returns the negated
+.IR errno
+for operations submitted to the ring. Instead it returns the negated
 .I errno
 directly in the CQE
 .I res
--- a/man/io_uring_prep_fallocate.3
+++ b/man/io_uring_prep_fallocate.3
@@ -47,8 +47,8 @@ details on possible values. Note that where synchronous system calls will return
 on failure and set
 .I errno
 to the actual error value, io_uring never uses
-.IR errno .
-Instead it returns the negated
+.IR errno
+for operations submitted to the ring. Instead it returns the negated
 .I errno
 directly in the CQE
 .I res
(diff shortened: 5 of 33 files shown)


ammarfaizi2:
> "io_uring never uses `errno`" is too strong; io_uring syscalls set `errno`.

No, they don't set `errno`.

`errno` is from libc and we don't use libc for x86, x86-64, and aarch64.

calebsander:
> No, they don't set errno.

The man pages, e.g. `man/io_uring_enter.2` seem to suggest otherwise:
> **RETURN VALUE**
>       ...
>       Errors that occur not on behalf of a submission queue entry are returned via the system call directly. On such an error, -1 is returned and errno is set appropriately.

ammarfaizi2:
> The man pages, e.g. `man/io_uring_enter.2` seem to suggest otherwise

We should fix this man page to tell the users that liburing doesn't touch `errno` from libc.

ammarfaizi2:
> > The man pages, e.g. `man/io_uring_enter.2` seem to suggest otherwise
> 
> We should fix this man page to tell the users that liburing doesn't touch `errno` from libc.

Can you do that?

calebsander:
`io_uring_enter()` is not a liburing function (the same is true of `io_uring_setup()` and `io_uring_register()`). These are syscalls, implemented by the kernel. I believe the kernel _does_ set `errno` when these syscalls fail (which is how other syscalls behave), so the man pages are correct.

ammarfaizi2:
> `io_uring_enter()` is not a liburing function (the same is true of `io_uring_setup()` and `io_uring_register()`).

Where have they actually been implemented that set `errno`?

> I believe the kernel _does_ set `errno` when these syscalls fail

`errno` is a userspace thing, the kernel doesn't know about `errno`.
The kernel simply returns a value within the range `[-4095, -1]` to
indicate that the syscall fails. This is why we have:

https://github.com/axboe/liburing/blob/b1486931117f643b0ad6612beb3f9563b508126c/src/syscall.h#L68-L71

calebsander:
I guess I have been using liburing 2.1. Seems like liburing has changed since then to no longer make the syscalls through libc, which I assume is what was setting `errno` when I tested it. I can't actually figure out how to call the syscalls directly; the suggested `linux/io_uring.h` include from the man page doesn't seem to declare any functions:
```
       #include <linux/io_uring.h>

       int io_uring_enter(unsigned int fd, unsigned int to_submit,
                          unsigned int min_complete, unsigned int flags,
                          sigset_t *sig);
```

ammarfaizi2:
@calebsander I posted an RFC patchset to implement those functions and fix the manpage.

https://lore.kernel.org/io-uring/20220829030521.3373516-1-ammar.faizi@intel.com/T/

Comments welcome...

PR #649, opened 2022-08-29
[GIT PULL] queue: assume direct SQ-SQE mapping

calebsander:
`_io_uring_get_sqe()` assumes SQ and SQE indices are interchangeable,
as it checks whether `sqe_tail - sq_head <= ring_entries`.
Under this assumption, we can set the SQ array during initialization
instead of every time SQEs are submitted.
We can also avoid dereferencing `ktail`, since `*ktail == sqe_head`.

----
## git request-pull output:
```
The following changes since commit bf248850dc2ae45d29d4fdde688e90d24f3dd6d2:

  Merge branch 'fix/flush-sq-unsigned' of https://github.com/calebsander/liburing (2022-08-28 19:27:44 -0600)

  git@github.com:calebsander/liburing.git refactor/sq-init

for you to fetch changes up to c65a873711d74d521f24a914d23653985bc7fa6d:

  queue: assume direct SQ-SQE mapping (2022-08-29 11:07:00 -0600)

Caleb Sander (1):
      queue: assume direct SQ-SQE mapping

 src/queue.c | 33 ++++++++++-----------------------
 src/setup.c | 10 ++++++++++
 2 files changed, 20 insertions(+), 23 deletions(-)
```
----

### DIFF
--- a/src/queue.c
+++ b/src/queue.c
@@ -184,28 +184,15 @@ done:
 unsigned __io_uring_flush_sq(struct io_uring *ring)
 {
 	struct io_uring_sq *sq = &ring->sq;
-	const unsigned mask = sq->ring_mask;
-	unsigned ktail = *sq->ktail;
-	unsigned to_submit = sq->sqe_tail - sq->sqe_head;
+	unsigned tail = sq->sqe_tail;
 
-	if (!to_submit)
-		goto out;
-
-	/*
-	 * Fill in sqes that we have queued up, adding them to the kernel ring
-	 */
-	do {
-		sq->array[ktail & mask] = sq->sqe_head & mask;
-		ktail++;
-		sq->sqe_head++;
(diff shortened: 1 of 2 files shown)


axboe:
Good point, liburing never uses the index differently than that. Some raw io_uring users, like fio, do. But with the liburing API, that's not really an issue.

calebsander:
Yes, liburing already breaks if users try to interchange setting `sq->array` manually and using `io_uring_get_sqe()`/`io_uring_submit()`. (I think it is possible to support this use case if we replace the `sqe_tail - sq_head` calculation with `sqe_tail - sqe_head + sq_tail - sq_head`. But I don't really see a use case for this.)
Initializing `sq->array` up front has a small cost for users who are going to do it manually, but seems like it could provide a reasonable steady-state performance improvement when submitting via liburing.

PR #698, opened 2022-10-20
Clarify that user_data is neither set nor cleared

ioquatix:
What do you think of being a little bit more explicit here?


axboe:
Yep that looks good to me, that's a better way to say it.

Issue #708, opened 2022-10-24
build broken on systems without linux/time_types.h


metze-samba:
Commit 00dcfc1f910fd8e191bf758b0a9b1d9f291460f5 ("io_uring: uapi: Sync with the kernel") broke
the build on systems without linux/time_types.h.

configure has code to cope a missing time_types.h, but the kernel's uapi header
forces it currently...

PR #709, opened 2022-10-24
uapi: Sync with the kernel in order to build systems without linux/time_types.h

metze-samba:
Fixes: https://github.com/axboe/liburing/issues/708


axboe:
I never do blind imports, so there's no risk of that. I'll sync them up on the kernel side next time I need to do an update anyway.
Oh, and hadn't noticed vger having issues, I did get your email. I'll reply there.

BTW, this one has a typo in it, but it's inherited from the kernel side. So that should get fixed up there...

Issue #660, opened 2022-09-22
Is there a technical reason why some of the header definitions are "static inline" and so don't have exported symbols?


GavinRay97:
I've been trying to integrate `liburing` with the JVM's new Panama functionality, which allows you to do FFI and interop with foreign memory without writing bindings/stubs.

The major issue I've run into is that the majority of the API seems to be defined as `static inline`, so when I try to call them, they simply don't exist:

- https://github.com/GavinRay97/panama-liburing/blob/82f83a247bb78c26d787992fca28ea1b951dfd3c/lib/src/test/java/panama/liburing/LibraryTest.java#L38-L44

```java
SegmentAllocator allocator = SegmentAllocator.newNativeArena(session);

// Works
MemorySegment ring = io.uring.io_uring.allocate(allocator);
io_uring_queue_init(QUEUE_DEPTH, ring, 0);

// Works
MemoryAddress sqe = io_uring_get_sqe(ring);

// ERROR HERE:
//
// java.lang.UnsatisfiedLinkError: unresolved symbol: io_uring_prep_rw
io_uring_prep_rw(/* op */ io.uring.liburing_h.IORING_OP_NOP(),
    /* sqe */ sqe, /* fd */ 0, /* addr */ MemoryAddress.NULL, /* len */ 0, /* offset */ 0);
```

I don't know much about C/systems programming so I thought there might be a good reason why they were like this.

If the changes can't be made repo-wide, is there any detriment to applying a patch on top of a personal library that removes the `static inline` from the definitions?

Thank you =)

axboe:
If it's in a header, it needs to be static inline or you'll end up with duplicates. And the reason they are static inline rather than an exported symbol is to avoid the function call overhead for them.

For a private library, you can move them into C files for example and just export them. It'll still work fine, no functionality depends on them being inline.

notinaboat:
@axboe I ran into the same issue calling `liburing` from Julia.

I am guessing that the same issue will effect each new language that wants to interface with `liburing`.

Would you consider adding a C file to the library with functions something like `_noinline__io_uring_prep_poll_add()` to wrap each of the `static inline` functions ?

The alternatives are:
 1. Each language interface library builds its own C wrapper for the `static inline` functions
 2. Each language interface library translates the content of the `static inline` functions into the target language.

Either of these alternatives increase the risk of errors due to repetition of work.
[...]

PR #743, opened 2022-12-02
[GIT PULL] Liburing FFI

cmazakas:
Add compiled versions of all the `static inline` functions in liburing.h for easier FFI interoperability with languages like Rust.

----
## git request-pull output:
```
The following changes since commit 5772879454421e82ae153c60853cf378a314e8f9:

  Merge branch 'provide-buffers-man-fixes' of https://github.com/wlukowicz/liburing (2022-12-03 09:27:37 -0700)

  https://github.com/cmazakas/liburing.git liburing-ffi

for you to fetch changes up to 9e2890d35e9677d8cfc7ac66cdb2d97c48a0b5a2:

  build: add liburing-ffi (2022-12-05 20:35:12 -0800)

Christian Mazakas (1):
      build: add liburing-ffi

 .gitignore             |   2 ++
 src/Makefile           |  21 +++++++++++++++++++-
 src/ffi.c              |  14 +++++++++++++
 src/include/liburing.h | 222 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++---------------------------------------------------------------------------------------------------
 src/liburing-ffi.map   | 165 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 5 files changed, 316 insertions(+), 108 deletions(-)
 create mode 100644 src/ffi.c
 create mode 100644 src/liburing-ffi.map
```

----


axboe:
@cmazakas There's a build failure, can you take a look?

cmazakas:
> @cmazakas There's a build failure, can you take a look?

Seems like clang isn't super stoked on my Big Idea of putting definitions in the header without inline or static.

The easiest thing to do is to simply disable the warning via a pragma. Other options for generating compiled versions of the liburing API are more onerous.

notinaboat:
Looks like `-Wmissing-prototypes` was only recently added : https://github.com/axboe/liburing/commit/fd6b571b0b03aeeb529f235c5c9c0a7c3256340c#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721L29-L30

Need something like this (with a layer of `#ifdef __clang__`).
```C
#define LIB_URING_STATIC_INLINE
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wmissing-prototypes"
#include "liburing.h"
#pragma clang diagnostic pop
```

src/include/liburing.h
```diff
@@ -34,6 +34,10 @@
 #define uring_likely(cond)	__builtin_expect(!!(cond), 1)
 #endif
 
+#ifndef LIB_URING_STATIC_INLINE
+#define LIB_URING_STATIC_INLINE static inline
```
axboe, line 38: Can we calls this something a bit more condensed? It looks like just search/replace was done, which means we now have lines that are > 80 chars and the careful aligning that was otherwise done now just looks like a mess.

src/include/liburing.h
```diff
@@ -152,7 +156,7 @@ struct io_uring_probe *io_uring_get_probe(void);
  */
 void io_uring_free_probe(struct io_uring_probe *probe);
 
-static inline int io_uring_opcode_supported(const struct io_uring_probe *p,
+LIB_URING_STATIC_INLINE int io_uring_opcode_supported(const struct io_uring_probe *p,
```
axboe, line 159: The very first one is a good example - first line now is too long, and the 2nd line is just lazily hanging now with deprecated spacing.

src/liburing-ffi.map
```diff
@@ -0,0 +1,180 @@
+LIBURING_2.0 {
```
axboe, line 1: Why not just make this for 2.4 and onwards? The older release are not being redone, so not sure why we'd want to have anything but 2.4?

axboe:
Maybe this would end up being cleaner if a prep patch split all the prep helpers from liburing.h into a separate header, which you could then use from both liburing.h and for the exported case too?

src/liburing-ffi.map
cmazakas: You kinda caught me here, I must admit.

I've actually never written a version script before. To get something working, I simply copied what already existed.

You are correct, any version before 2.4 is nonsensical in this instance.

src/include/liburing.h
cmazakas: I apologize for this.

I'm used to more verbose macros.

We wouldn't need to do anything if we simply used `STATIC_INLINE`. Same characters and length.

This also reminds me, I didn't `#undef` the macro at the end of the header. Not best practices.

cmazakas:
> Maybe this would end up being cleaner if a prep patch split all the prep helpers from liburing.h into a separate header, which you could then use from both liburing.h and for the exported case too?

It's not just the prep helpers, but anything else `static inline` that a user might want. This essentially winds up being the entire liburing.h header. Maybe I'm misunderstanding though. In all honesty, I can't say that I'm overwhelmingly familiar with liburing. I chose Rust as my language to learn the library in the first place.

I think given the notes above, we can make things work nicely by just using `STATIC_INLINE` as our macro, remembering to `#undef` it at the end of the file. Or any other 13 character macro name. For example, `IOURING_INLNE`.

I also forgot to update the .gitignore as well, it seems.

src/include/liburing.h
axboe: Both of those sound good.

GavinRay97:
This is super useful! It rids the need of the hacks I've maintained to use `liburing` in D and other languages

Question I asked a while ago:
https://github.com/axboe/liburing/issues/660

cmazakas:
> Out of curiosity, why does it generate a separate `liburing-ffi.a` library?

I didn't want to penalize current users of liburing.h who benefit from everything being a `static inline` function by ballooning their binaries. So, this way, C and C++ users can just `#include <liburing.h>` and link against liburing.a like they could before while Rust, D, etc. users can simply just link against liburing-ffi.a.

I also had concerns about multiple symbols coming into play during the link phase.

GavinRay97:
> static inline function by ballooning their binaries

Ah that makes sense, since I suppose that the default installation is what goes into distro package managers and whatnot, when you do like `dpkg install liburing`

cmazakas:
@axboe Alright, I finally remembered to update the `git request-pull` in the top comment. I think this should be all good to go unless there's anything else you'd like to see!

ammarfaizi2:
> I think this should be all good to go unless there's anything else you'd like to see!

Why are these
```
	io_uring_major_version;
	io_uring_minor_version;
	io_uring_check_version;
```
not included in the liburing-ffi.map?

axboe:
From a quick diff, does look like those are the only 3 missing. @cmazakas care to do a quick patch for those?

ammarfaizi2:
> care to do a quick patch for those?

```patch
From: Ammar Faizi <ammarfaizi2@gnuweeb.org>
Date: Thu, 8 Dec 2022 00:30:55 +0700
Subject: [PATCH] liburing-ffi.map: Add missing functions

Add these 3 missing functions:

  io_uring_major_version;
  io_uring_minor_version;
  io_uring_check_version;

Link: https://github.com/axboe/liburing/pull/743#issuecomment-1341310572
Fixes: 9e2890d35e9677d8cfc7ac66cdb2d97c48a0b5a2 ("build: add liburing-ffi")
Signed-off-by: Ammar Faizi <ammarfaizi2@gnuweeb.org>
---
 src/liburing-ffi.map | 4 ++++
 1 file changed, 4 insertions(+)

diff --git a/src/liburing-ffi.map b/src/liburing-ffi.map
index 71834cd..8dd4641 100644
--- a/src/liburing-ffi.map
+++ b/src/liburing-ffi.map
@@ -56,6 +56,10 @@ LIBURING_2.4 {
 		io_uring_get_events;
 		io_uring_submit_and_get_events;
 
+		io_uring_major_version;
+		io_uring_minor_version;
+		io_uring_check_version;
+
 		io_uring_peek_cqe;
 		io_uring_prep_timeout_update;
 		io_uring_buf_ring_init;
-- 
Ammar Faizi

```

GavinRay97:
Thanks all! 🥳 
Big step in `io_uring`'s ease of interop + consumption from other languages, great to have this small-but-significant PR =)

PR #748, opened 2022-12-05
WIP: liburing, man: make io_uring_sqring_wait() more consistent with API

dmantipov:
Make `io_uring_sqring_wait()` return an amount of free SQ ring entries or `-EINVAL` if the ring was not setup with `IORING_SETUP_SQPOLL`, adjust manual page accordingly.

Signed-off-by: Dmitry Antipov <dantipov@cloudlinux.com>


src/include/liburing.h
```diff
@@ -1093,20 +1093,21 @@ static inline unsigned io_uring_sq_space_left(const struct io_uring *ring)
...
 static inline int io_uring_sqring_wait(struct io_uring *ring)
 {
-	if (!(ring->flags & IORING_SETUP_SQPOLL))
-		return 0;
-	if (io_uring_sq_space_left(ring))
-		return 0;
+	int ret;
 
-	return __io_uring_sqring_wait(ring);
+	if (!(ring->flags & IORING_SETUP_SQPOLL))
```
axboe, line 1107: This does make it follow the documentation, but could also potentially break existing use cases. However, I don't think the risk of that is high, so probably fine to do this.

src/include/liburing.h
```diff
@@ -1093,20 +1093,21 @@ static inline unsigned io_uring_sq_space_left(const struct io_uring *ring)
...
+		return -EINVAL;
+	return (ret = io_uring_sq_space_left(ring))
```
axboe, line 1109: This is unreadable, please write this without using a ternary. It's made worse by the fact that you double the assignment as the value as well.

And if re-written a bit, it also means you don't need a kernel change.

src/include/liburing.h
dmantipov: Maybe any particular use case you're aware of?

src/include/liburing.h
dmantipov: __it also means you don't need a kernel change__

Are you sure? I didn't dig the kernel part too much but it seems that `io_uring_enter(fd, 0, 0,  IORING_ENTER_SQ_WAIT, NULL)` will return 0 without this change.

src/include/liburing.h
axboe: As mentioned, not aware of any problematic cases. Just raising it as a potential one, as it is a change in return value.

src/include/liburing.h
axboe: If the kernel side returns 0, you'd use the helper to return available entries. You don't need the kernel to return this for you.

src/include/liburing.h
```diff
@@ -1093,18 +1093,23 @@ static inline unsigned io_uring_sq_space_left(const struct io_uring *ring)
...
-		return 0;
-	if (io_uring_sq_space_left(ring))
-		return 0;
+		return -EINVAL;
+
+	ret = io_uring_sq_space_left(ring);
+	if (ret)
+		return ret;
 
 	return __io_uring_sqring_wait(ring);
```
axboe, line 1114: You'd need something ala:

```
ret = io_uring_sq_space_left(ring);
if (ret)
    return ret;

ret = __io_uring_sqring_wait(ring);
if (ret < 0)
    return ret;

return io_uring_sq_space_left(ring);
```

or something like that.

src/include/liburing.h
dmantipov: OK but this requires an extra call to helper (which hopefully may be avoided with just one-line kernel change).

PR #765, opened 2023-01-03
[GIT PULL] Add support for io_uring_prep_msg_ring_cqe_flags(3)

leitao:
This PR adds support for a new flag to pass through flags from SQE to CQE. 

Basically, you specify the flags you want to see in `cqe->flags` by setting `sqe->file_index`, and toggling the new flag `IORING_MSG_RING_FLAGS_PASS` 

Kernel patch: https://lore.kernel.org/io-uring/20230103160507.617416-1-leitao@debian.org/T/#u

----
## git request-pull output:
```
The following changes since commit 052c2e03e4b64e07f9d4e43ed816cd8e5721b9d8:

  Merge branch 'io_uring_uapi' of https://github.com/metze-samba/liburing (2022-12-28 10:04:10 -0700)

  git@github.com:leitao/liburing.git master

for you to fetch changes up to f08da78a3b0ff6c175bbaeb8ddd35cb11a1635d8:

  test/io_uring_prep_msg_ring_cqe_flags new test (2023-01-03 06:51:17 -0800)

Breno Leitao (4):
      uapi: Sync with the kernel in order to see IORING_MSG_RING_FLAGS_PASS
      Add io_uring_prep_msg_ring_cqe_flags function
      man/io_uring_prep_msg_ring: Add additional function
      test/io_uring_prep_msg_ring_cqe_flags new test

 man/io_uring_prep_msg_ring.3           |  18 +++++++++++
 man/io_uring_prep_msg_ring_cqe_flags.3 |   1 +
 src/include/liburing.h                 |   9 ++++++
 src/include/liburing/io_uring.h        |   2 ++
 test/Makefile                          |   1 +
 test/msg-ring-flags.c                  | 120 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 6 files changed, 151 insertions(+)
 create mode 120000 man/io_uring_prep_msg_ring_cqe_flags.3
 create mode 100644 test/msg-ring-flags.c
```

### DIFF
--- a/man/io_uring_prep_msg_ring.3
+++ b/man/io_uring_prep_msg_ring.3
@@ -14,6 +14,13 @@ io_uring_prep_msg_ring \- send a message to another ring
 .BI "                        unsigned int " len ","
 .BI "                        __u64 " data ","
 .BI "                        unsigned int " flags ");"
+.PP
+.BI "void io_uring_prep_msg_ring_cqe_flags(struct io_uring_sqe *" sqe ","
+.BI "                        int " fd ","
+.BI "                        unsigned int " len ","
+.BI "                        __u64 " data ","
+.BI "                        unsigned int " flags ","
+.BI "                        unsigned int " cqe_flags ");"
 .fi
 .SH DESCRIPTION
 .PP
@@ -46,6 +53,18 @@ and
 fields. The use case may be anything from simply waking up someone waiting
 on the targeted ring, or it can be used to pass messages between the two
 rings.
+
+.BR io_uring_prep_msg_ring_cqe_flags (3)
+is similar to
+.BR io_uring_prep_msg_ring (3) .
+But has an addition
+.I cqe_flags
+parameter, which is used to set
+.I flags
+field on CQE side. That way, you can set the CQE flags field
+.I cqe->flags
+when sending a message. Be aware that io_uring could potentially set additional
+bits into this field.
 
 .SH RETURN VALUE
 None
--- a/src/include/liburing.h
+++ b/src/include/liburing.h
@@ -987,6 +987,15 @@ IOURINGINLINE void io_uring_prep_link(struct io_uring_sqe *sqe,
 	io_uring_prep_linkat(sqe, AT_FDCWD, oldpath, AT_FDCWD, newpath, flags);
 }
 
+IOURINGINLINE void io_uring_prep_msg_ring_cqe_flags(struct io_uring_sqe *sqe,
+					  int fd, unsigned int len, __u64 data,
+					  unsigned int flags, unsigned int cqe_flags)
+{
+	io_uring_prep_rw(IORING_OP_MSG_RING, sqe, fd, NULL, len, data);
+	sqe->msg_ring_flags = IORING_MSG_RING_FLAGS_PASS | flags;
+	sqe->file_index = cqe_flags;
+}
+
 IOURINGINLINE void io_uring_prep_msg_ring(struct io_uring_sqe *sqe, int fd,
 					  unsigned int len, __u64 data,
 					  unsigned int flags)
--- a/test/msg-ring-flags.c
+++ b/test/msg-ring-flags.c
@@ -1,0 +1,131 @@
+/* SPDX-License-Identifier: MIT */
+/*
+ * Description: test ring messaging with flags command
+ *
+ */
+#include <errno.h>
+#include <stdio.h>
+#include <unistd.h>
+#include <stdlib.h>
+#include <string.h>
+#include <fcntl.h>
+#include <pthread.h>
+
+#include "liburing.h"
+#include "helpers.h"
+
+#define CUSTOM_FLAG 0x42
+#define USER_DATA 0x5aa5
+#define LEN 0x20
+#define ID 0x1
+
+static int recv_msg(struct io_uring *ring)
+{
+	struct io_uring_cqe *cqe;
+	int ret;
+
+	ret = io_uring_wait_cqe(ring, &cqe);
+	if (ret) {
+		fprintf(stderr, "wait cqe %d\n", ret);
+		return T_EXIT_FAIL;
+	}
+	if (cqe->user_data != USER_DATA) {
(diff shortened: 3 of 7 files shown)


src/include/liburing.h
```diff
@@ -987,6 +987,15 @@ IOURINGINLINE void io_uring_prep_link(struct io_uring_sqe *sqe,
 	io_uring_prep_linkat(sqe, AT_FDCWD, oldpath, AT_FDCWD, newpath, flags);
 }
 
+IOURINGINLINE void io_uring_prep_msg_ring_cqe_flags(struct io_uring_sqe *sqe,
+					  int fd, unsigned int len, __u64 data,
+					  unsigned int flags, unsigned int cqe_flags)
+{
+	io_uring_prep_rw(IORING_OP_MSG_RING, sqe, fd, NULL, len, data);
+	sqe->rw_flags = IORING_MSG_RING_FLAGS_PASS | flags;
```
axboe, line 995: Should probably use ->msg_ring_flags

test/msg-ring-flags.c
```diff
@@ -0,0 +1,120 @@
...
+
+#include "liburing.h"
+#include "helpers.h"
+
+#define CUSTOM_FLAG 0x42
+#define USER_DATA 0x5aa5
+#define LEN 0x20
+#define ID 0x1
+
+static int recv_msg(struct io_uring *ring)
```
axboe, line 22: Would be nice if this and send_msg() used T_EXIT_PASS/T_EXIT_FAIL as return values rather than assume 0/1 are the same.

axboe:
Don't see anything immediately wrong here, but we need the kernel side sorted first before anything can go into liburing.

src/include/liburing.h
leitao: Thanks. Should we do the same for `io_uring_prep_msg_ring()`, which is using `        sqe->rw_flags` as well?

```
IOURINGINLINE void io_uring_prep_msg_ring(struct io_uring_sqe *sqe, int fd,
                                          unsigned int len, __u64 data,
                                          unsigned int flags)
{
        io_uring_prep_rw(IORING_OP_MSG_RING, sqe, fd, NULL, len, data);
        sqe->rw_flags = flags;
}

```

ammarfaizi2 (commented):
Can you update the CHANGELOG file and mention this new function?

I think the CHANGELOG change can also be folded into
"Add io_uring_prep_msg_ring_cqe_flags function" patch.

leitao:
> Can you update the CHANGELOG file and mention this new function?

Sure. Done now: https://github.com/axboe/liburing/pull/765/commits/27180d7be0590dce0f88f0ca94ab5e3d1c1e1a3f

man/io_uring_prep_msg_ring.3
```diff
@@ -47,6 +54,17 @@ fields. The use case may be anything from simply waking up someone waiting
 on the targeted ring, or it can be used to pass messages between the two
 rings.
 
+.BR io_uring_prep_msg_ring_cqe_flags(3)
```
axboe, line 57: Need a space after flags and (3)

man/io_uring_prep_msg_ring.3
```diff
@@ -47,6 +54,17 @@ fields. The use case may be anything from simply waking up someone waiting
...
+is similar to
+.BR io_uring_prep_msg_ring(3) .
```
axboe, line 59: Same space missing here.

man/io_uring_prep_msg_ring.3
```diff
@@ -47,6 +54,17 @@ fields. The use case may be anything from simply waking up someone waiting
...
+But requests an additional parameter
```
axboe, line 60: "But has an addition
.I cqe_flags
parameter, which is used to set [...]"

man/io_uring_prep_msg_ring.3
```diff
@@ -47,6 +54,17 @@ fields. The use case may be anything from simply waking up someone waiting
...
+.I cqe_flags
+which is used to set the
+.I flags
+field at CQE side. That way, you can set the CQE flags field
```
axboe, line 64: on the CQE side

axboe: Are these addition flags or the only flags? I think that should be clarified.

leitao: Sorry @axboe , I am not sure what you meant by this last sentence. What do you mean by "addition flags or the only flags"?

src/include/liburing.h
leitao: https://github.com/axboe/liburing/pull/771 contains the fix.

test/msg-ring-flags.c
```diff
@@ -0,0 +1,120 @@
...
+		fprintf(stderr, "sqe submit failed: %d\n", ret);
+		return ret;
+	}
+
+	ret = io_uring_wait_cqe(ring, &cqe);
+	if (ret < 0) {
+		fprintf(stderr, "wait completion %d\n", ret);
+		return ret;
+	}
+	if (cqe->res != 0) {
```
axboe, line 75: What happens if this is run on a kernel that doesn't support this new feature, or even OP_MSG_RING? We should catch that case and ensure that T_EXIT_SKIP is returned for that case rather than trigger a failure. See what lots of the other feature based test cases do for that.

man/io_uring_prep_msg_ring.3
leitao: Thanks for the clarification. I've added an extra line to clarify that `io_uring` could potentially set additional bits in this field.

test/msg-ring-flags.c
leitao: This is a good point. I am returning `T_EXIT_SKIP` in two cases now:

1) if `io_uring_submit()` return -EINVAL
2) if `cqe->res == -EINVAL` after `io_uring_wait_cqe(ring, &cqe)`. Let me know if this is not enough.

axboe:
Thanks, merged with a cleanup on top.

PR #925, opened 2023-08-20
[GIT PULL] update io_uring_prep_close.3

cmazakas:
Setting the `IOSQE_FIXED_FILE` flag in the direct close() SQE causes erroneous behavior users may not be aware of.

Update the docs on `io_uring_prep_close()` to specify if the `IOSQE_FIXED_FILE` flag should be set or not and what it's ramifications are.
----
## git request-pull output:
```
The following changes since commit 660cefae669c985aeaeaf50e45b635777ac40edc:

  test/wq-aff: add test case for SQPOLL io-wq affinity (2023-08-14 08:21:02 -0600)

  https://github.com/cmazakas/liburing.git io_uring_prep_close_man_page_updates

for you to fetch changes up to be521645cad1dcc22b3909cad1662c139afd5e55:

  update io_uring_prep_close.3 (2023-08-20 12:50:44 -0700)

Christian Mazakas (1):
      update io_uring_prep_close.3

 man/io_uring_prep_close.3 | 10 ++++++++++
 1 file changed, 10 insertions(+)
```
----
<details>

## Pull Request Guidelines
```
```

2. Then an empty line.
```
```

certain specific format).

of truth, not github.

URL.

```
```

Use a real email address!

### Commit message example:
```

safe without synchronization.

```

</details>

----

### DIFF
--- a/man/io_uring_prep_close.3
+++ b/man/io_uring_prep_close.3
@@ -32,6 +32,16 @@ argument instead of the
 This is identical to unregistering the direct descriptor, and is provided as
 a convenience.
 
+To properly close a direct descriptor, the 
+.I sqe
+flag
+.B IOSQE_FIXED_FILE
+should not be set. If the flag is set, the descriptor can be reused after a
+manual call to
+.BR io_uring_register_files_update (3) .
+Otherwise the direct descriptor will not be available to operations that use
+.BR IORING_FILE_INDEX_ALLOC .
+
 These functions prepare an async
 .BR close (2)
 request. See that man page for details.


axboe:
I think this would be a lot more useful if it explains how the sqe is setup for a direct close, rather than just detail one way it definitely doesn't work. Under the covers, they basically all use ```__io_uring_set_target_fixed_file()``` which sets sqe->file_index. This is different than a file that just takes a direct descriptor without manipulating it, where you would indeed use ```IOSQE_FIXED_FILE``` to tell the kernel that the ```sqe->fd``` is a direct descriptor index rather than a normal file descriptor.

cmazakas:
You know what's interesting, I actually have no idea when, where or why we're supposed to set `IOSQE_FIXED_FILE`.

I tried looking into the liburing src and more of the docs and I feel like I have more questions than answers now.

Where do I read more about this?

axboe:
See the io_uring_enter.2 man page, it has a description of the sqe flags. But it's pretty simple - if your request takes a file descriptor, then you can set this flag and then io_uring will interpret it as "fd isn't a normal file descriptor, it's a direct descriptor".

cmazakas:
Alright, so maybe something like...

_To properly close a direct descriptor, perform something like the following:_
```cpp
struct io_uring_sqe *sqe = io_uring_get_sqe(ring);
io_uring_prep_close_direct(sqe, fd);
io_uring_sqe_set_flags(sqe, IOSQE_CQE_SKIP_SUCCESS);
io_uring_sqe_data(sqe, NULL);
io_uring_submit(ring);
```

PR #1004, opened 2023-11-24
[GIT PULL] Add ftruncate

tontinton:
To accompany the patches I sent for adding ftruncate, here are the changes for usermode.

----
## git request-pull output:
```
The following changes since commit 63342497bec87b72d4b415c8bed930c15ff83b0d:

  test/read-mshot: check that IORING_CQE_F_MORE isn't set on error (2024-01-22 14:56:08 -0700)

  git@github.com:tontinton/liburing.git truncate

for you to fetch changes up to c023393f3a3b960e3df77e5bcfd1b2e93e3b5e0f:

  man/io_uring_prep_ftruncate: Add the new ftruncate command (2024-01-31 17:52:13 +0200)

Tony Solomonik (4):
      Add ftruncate helpers
      test/truncate: Add test for ftruncate
      test/truncate: Add test for failure on truncate path
      man/io_uring_prep_ftruncate: Add the new ftruncate command

 man/io_uring_prep_ftruncate.3   |  48 ++++++++++++++++++++++++++++++++
 src/include/liburing.h          |   6 ++++
 src/include/liburing/io_uring.h |   1 +
 src/liburing-ffi.map            |   1 +
 test/Makefile                   |   1 +
 test/truncate.c                 | 179 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 6 files changed, 236 insertions(+)
 create mode 100644 man/io_uring_prep_ftruncate.3
 create mode 100644 test/truncate.c
```
----
<details>

## Pull Request Guidelines
```
```

2. Then an empty line.
```
```

certain specific format).

of truth, not github.

URL.

```
```

Use a real email address!

### Commit message example:
```

safe without synchronization.

```

</details>

----

### DIFF
--- a/src/liburing-ffi.map
+++ b/src/liburing-ffi.map
@@ -191,4 +191,5 @@ LIBURING_2.6 {
 	global:
 		io_uring_prep_fixed_fd_install;
 		io_uring_buf_ring_available;
+		io_uring_prep_ftruncate;
 } LIBURING_2.5;
--- a/test/truncate.c
+++ b/test/truncate.c
@@ -19,6 +19,44 @@
 #define ONE_GIG_SIZE ((loff_t)1024 * 1024 * 1024)
 #define HALF_GIG_SIZE ((loff_t)512 * 1024 * 1024)
 
+static int test_truncate(struct io_uring *ring, int fd)
+{
+	struct io_uring_cqe *cqe;
+	struct io_uring_sqe *sqe;
+	int ret = -1;
+
+	sqe = io_uring_get_sqe(ring);
+	if (!sqe) {
+		fprintf(stderr, "get sqe failed\n");
+		goto err;
+	}
+
+	memset(sqe, 0, sizeof(*sqe));
+
+	io_uring_prep_rw(IORING_OP_FTRUNCATE, sqe, fd, "fail", 0, 4);
+
+	ret = io_uring_submit(ring);
+	if (ret <= 0) {
+		fprintf(stderr, "sqe submit failed: %d\n", ret);
+		goto err;
+	}
+
+	ret = io_uring_wait_cqe(ring, &cqe);
+	if (ret < 0) {
+		fprintf(stderr, "wait completion %d\n", ret);
+		goto err;
+	}
+	ret = cqe->res;
+	io_uring_cqe_seen(ring, cqe);
+	if (ret != -EINVAL) {
+		fprintf(stderr, "unexpected truncate res %d\n", ret);
+		goto err;
+	}
+
+err:
+	return ret;
+}
+
 static int test_ftruncate(struct io_uring *ring, int fd, loff_t len)
 {
 	struct io_uring_cqe *cqe;
@@ -125,6 +163,10 @@ int main(int argc, char *argv[])
 			goto err;
 		}
 	}
+
+	ret = test_truncate(&ring, fd);
+	if (ret < 0)
+		goto err;
 
 out:
 	close(fd);
--- a/test/truncate.c
+++ b/test/truncate.c
@@ -1,0 +1,135 @@
+/* SPDX-License-Identifier: MIT */
+/*
+ * Description: run various truncate tests
+ *
+ */
+#include <errno.h>
+#include <stdio.h>
+#include <unistd.h>
+#include <stdlib.h>
+#include <string.h>
+#include <fcntl.h>
+#include <sys/stat.h>
+#include <sys/ioctl.h>
+
+#include "liburing.h"
+#include "helpers.h"
+
+#define TWO_GIG_SIZE ((loff_t)2 * 1024 * 1024 * 1024)
+#define ONE_GIG_SIZE ((loff_t)1024 * 1024 * 1024)
+#define HALF_GIG_SIZE ((loff_t)512 * 1024 * 1024)
+
+static int test_ftruncate(struct io_uring *ring, int fd, loff_t len)
+{
+	struct io_uring_cqe *cqe;
+	struct io_uring_sqe *sqe;
+	int ret;
+
+	sqe = io_uring_get_sqe(ring);
+	if (!sqe) {
+		fprintf(stderr, "get sqe failed\n");
+		goto err;
+	}
+
+	memset(sqe, 0, sizeof(*sqe));
+
+	io_uring_prep_ftruncate(sqe, fd, len);
+
+	ret = io_uring_submit(ring);
+	if (ret <= 0) {
+		fprintf(stderr, "sqe submit failed: %d\n", ret);
+		goto err;
+	}
+
+	ret = io_uring_wait_cqe(ring, &cqe);
+	if (ret < 0) {
+		fprintf(stderr, "wait completion %d\n", ret);
+		goto err;
+	}
+	ret = cqe->res;
+	io_uring_cqe_seen(ring, cqe);
+	return ret;
+err:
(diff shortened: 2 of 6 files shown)


axboe:
I think this looks fine, but one problem is that if you run this test case on an older kernel, then the test will fail. If you get -EINVAL for the _first_ ftruncate you try, just assume the kernel doesn't support it and return T_EXIT_SKIP for that and skip the other tests. That way it'll just skip the test case on older kernels, but test it on newer kernels.

See other test cases that do this very thing, usually with a ```static int no_ftruncate``` global that can be set by the first test invocation, and then checked again for non-zero when that first test case returns. If true, then it should return T_EXIT_SKIP.

tontinton:
It already runs `goto out` which returns 0, when ftruncate is not supported.

test/truncate.c
```diff
@@ -0,0 +1,130 @@
...
+	test_sizes[0] = TWO_GIG_SIZE;
+	test_sizes[1] = ONE_GIG_SIZE;
+	test_sizes[2] = HALF_GIG_SIZE;
+
+	for (i = 0; i < 3; i++) {
+		ret = test_ftruncate(&ring, fd, test_sizes[i]);
+		if (ret < 0) {
+			if (ret == -EBADF || ret == -EINVAL) {
+				fprintf(stdout, "Ftruncate not supported, skipping\n");
+				goto out;
```
axboe, line 150: Can be a break?

test/truncate.c
```diff
@@ -0,0 +1,130 @@
...
+		} else if (ret)
+			goto err;
+
+		if (get_file_size(fd, &size) || size != test_sizes[i]) {
+			fprintf(stderr, "size not %lu\n", size);
+			goto err;
+		}
+	}
+
+out:
```
axboe, line 167: Should return T_EXIT_PASS for success, T_EXIT_SKIP if the test is skipped, and T_EXIT_FAIL if the test case failed.

axboe:
> It already runs goto out which returns 0, when ftruncate is not supported.

It does, but it doesn't return T_EXIT_SKIPPED. I think you should also distinguish between "we got EBADF/EINVAL on the _first_ test, it's probably not supported, T_EXIT_SKIPPED" rather than just have any of the test cases returning EINVAL leading to a skipped test where it probably should've been a failure.

test/truncate.c
tontinton: If this is a break, it will go to test truncate with path failing, no reason to test that if ftruncate is unsupported I think

test/truncate.c
tontinton: Fixed

ammarfaizi2 (changes requested):
Missing Signed-off-by tag.

src/liburing-ffi.map
```diff
@@ -191,4 +191,5 @@ LIBURING_2.6 {
 	global:
 		io_uring_prep_fixed_fd_install;
 		io_uring_buf_ring_available;
+    io_uring_prep_ftruncate;
```
ammarfaizi2, line 194: Use tabs as indentation; don't use spaces.

test/truncate.c
```diff
@@ -19,6 +19,45 @@
...
+		goto err;
+	}
+	ret = cqe->res;
+	io_uring_cqe_seen(ring, cqe);
+	if (cqe->res != -EINVAL) {
+		fprintf(stderr, "unexpected truncate res %d\n", cqe->res);
+		goto err;
+	}
+
+  return 0;
```
ammarfaizi2, line 56: Same indentation issue.

ammarfaizi2, line 56: After this assignment, `ret = cqe->res,` the `ret` value is not used. Shouldn't it be returned?

Also, `goto err` is `return 1` while your caller checking is:
```
	ret = test_truncate(&ring, fd);
	if (ret < 0)
		goto err;
```
So `test_truncate()` will always be assumed as a success because it never returns a negative value.

ammarfaizi2 (changes requested):

test/truncate.c
tontinton: Fixed

src/liburing-ffi.map
tontinton: Fixed

test/truncate.c
tontinton: Fixed

test/truncate.c
```diff
@@ -19,6 +19,44 @@
...
+	}
+
+	ret = io_uring_wait_cqe(ring, &cqe);
+	if (ret < 0) {
+		fprintf(stderr, "wait completion %d\n", ret);
+		goto err;
+	}
+	ret = cqe->res;
+	io_uring_cqe_seen(ring, cqe);
+	if (cqe->res != -EINVAL) {
```
axboe, line 51: Doesn't matter too much here, but you cannot use the cqe after calling io_uring_cqe_seen() as it could've been overwritten by a new cqe. You assign 'ret' here anyway, so please just use ret. Would be a shame if people copied an error like that for their app.

tontinton: Fixed!

test/truncate.c
```diff
@@ -0,0 +1,135 @@
...
+{
+	struct io_uring_cqe *cqe;
+	struct io_uring_sqe *sqe;
+	int ret;
+
+	sqe = io_uring_get_sqe(ring);
+	if (!sqe) {
+		fprintf(stderr, "get sqe failed\n");
+		goto err;
+	}
```
ammarfaizi2, line 70: `goto err;` contains `return ret,` but `ret` is uninitialized; stop using `goto` if you don't have any cleanup. Just `return -1;`

test/truncate.c
```diff
@@ -0,0 +1,135 @@
...
+
+	memset(sqe, 0, sizeof(*sqe));
+
+	io_uring_prep_ftruncate(sqe, fd, len);
+
+	ret = io_uring_submit(ring);
+	if (ret <= 0) {
+		fprintf(stderr, "sqe submit failed: %d\n", ret);
+		goto err;
+	}
```
ammarfaizi2, line 80: If we assume `io_uring_submit()` returns 0 is failed, then you must override it with `return -1` because `return 0` is okay from the caller prespective.

ammarfaizi2 (changes requested):

test/truncate.c
ammarfaizi2: My mistake, it turns out `return 1;` Not `return ret`.

test/truncate.c
```diff
@@ -0,0 +1,135 @@
...
+	if (ret < 0) {
+		fprintf(stderr, "wait completion %d\n", ret);
+		goto err;
+	}
+	ret = cqe->res;
+	io_uring_cqe_seen(ring, cqe);
+	return ret;
+err:
+	return 1;
+}
```
ammarfaizi2, line 92: Return 1 is not an error as you have this in the caller:
```
	ret = test_ftruncate(&ring, fd, test_sizes[i]);
	if (ret < 0) {
		// fail
	}
```
Must be return -1?

ammarfaizi2: Ah no, it has an else if, my mistake again...

test/truncate.c
ammarfaizi2: Sorry for the noise, I probably read the old code before you force pushed.

PR #1216, opened 2024-09-01
[GIT PULL] man/io_uring_enter: add missing and clarify another return value

CPestka:
This makes three small changes to the man pages of io_uring_enter:
1. Adds documentation for the possible return value EEOWNERDEAD (i.e. ring uses sqpoll and some one killed the sqpoll kthread)
2. Adds a more specific description to when a thread is "invalid" and thus causes a return value of EEXIST. Which as far as i could tell boils down to IORING_ENTER_GETEVENTS && IORING_SETUP_DEFER_TASKRUN && current != ctx->submitter_task
3. Fixes a small grammatical error 

----
## git request-pull output:
```

The following changes since commit cc74667886e92a0a1b3a279ed5f2f527c5d9f3e0:

  Update CHANGELOG (2024-08-31 06:20:06 -0600)

  https://github.com/CPestka/liburing man_io_uring_enter

for you to fetch changes up to 6981007646a70326071ca3fa6d0b518ce0213ca4:

  man/io_uring_enter: Add missing return value EOWNERDEAD (2024-09-01 12:47:42 +0200)

CPestka (3):
      man/io_uring_enter: Fix minor grammatical error
      man/io_uring_enter: Clarify when EEXIST can occur
      man/io_uring_enter: Add missing return value EOWNERDEAD

 man/io_uring_enter.2 | 21 +++++++++++++++++++--
 1 file changed, 19 insertions(+), 2 deletions(-)

```
----
<details>

## Pull Request Guidelines
```
```
4. Follow the commit message format rules below.
5. Follow the Linux kernel coding style (see: https://github.com/torvalds/linux/blob/master/Documentation/process/coding-style.rst).

2. Then an empty line.
6. Then an empty line again (if it has a description).
7. Then a `Signed-off-by` tag with your real name and email. For example:
```
```

certain specific format).

of truth, not github.

URL.

```
```

Use a real email address!

### Commit message example:
```

safe without synchronization.

```

</details>

----

### DIFF
--- a/man/io_uring_enter.2
+++ b/man/io_uring_enter.2
@@ -44,12 +44,17 @@ number of events in
 before returning. This flag can be set along with
 .I to_submit
 to both submit and complete events in a single system call.
-If this flag is set the thread issuing the syscall must either be the
-thread that created the io_uring associated with
-.I fd
-or must not have the flag
+If this flag is set either the flag
 .B IORING_SETUP_DEFER_TASKRUN
-set.
+must not be set or the thread issuing the syscall must be the thread that
+created the io_uring associated with
+.I fd,
+or be the thread that enabled the ring originally created with
+.B IORING_SETUP_R_DISABLED
+via
+.BR io_uring_register (2)
+or
+.BR io_uring_enable_rings (3).
 .TP
 .B IORING_ENTER_SQ_WAKEUP
 If the ring has been created with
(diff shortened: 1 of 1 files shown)


man/io_uring_enter.2
```diff
@@ -44,6 +44,12 @@ number of events in
 before returning. This flag can be set along with
 .I to_submit
 to both submit and complete events in a single system call.
+If this flag is set the thread issuing the syscall must either be the
```
axboe, line 47: Minor clarification here - it is possible to have another thread create it, if it was created in a disabled state with ```IORING_SETUP_R_DISABLED```. Then it just requires the right thread to enable the ring with ```io_uring_enable_rings()```. So either one of those must be true:

1) Created by the thread that is going to use it, or
2) Created by another thread in a disabled state, and enabled by the thread that is going to use it

axboe:
Thanks!

PR #1224, opened 2024-09-06
configure: don't enable ublk if modern commands not available

Oipo:
If ublk_cmd.h does not contain modern command definitions such as UBLK_U_CMD_START_DEV the ublk io_uring commands cannot be used. These definitions have only been added last year, so there will be users who will not have these available.

Tested on my raspberry pi 4 with debian/bookworm and kernel 6.1.0-rpi7-rpi-v8

----
## git request-pull output:
```
<The following changes since commit f2dc904171f36f32a6f2566d1e6b12398ec3bb4f:

  man/io_uring_prep_close: note on direct descriptor failure cases (2024-09-05 09:00:54 -0600)

  git@github.com:Oipo/liburing.git 

for you to fetch changes up to 60ae198ac7dcdfc0580a354b203df092db81111c:

  configure: don't enable ublk if modern commands not available (2024-09-06 22:45:08 +0200)

Michael de Lang (1):
      configure: don't enable ublk if modern commands not available

 configure | 6 ++++++
 1 file changed, 6 insertions(+)
```
----
<details>

## Pull Request Guidelines
```
```

2. Then an empty line.
```
```

certain specific format).

of truth, not github.

URL.

```
```

Use a real email address!

### Commit message example:
```

safe without synchronization.

```

</details>

----

### DIFF
--- a/configure
+++ b/configure
@@ -480,6 +480,12 @@ cat > $TMPC << EOF
 int main(int argc, char **argv)
 {
   struct ublksrv_ctrl_cmd cmd = { };
+
+  struct ublk_ctrl_cmd_data data = {
+    .cmd_op	= UBLK_U_CMD_START_DEV,
+    .flags	= CTRL_CMD_HAS_DATA,
+  };
+
   return cmd.queue_id;
 }
 EOF


axboe:
Was too quick on this one, it's totally broken and can't ever work. It's mixing private structures and defines with public uapi ones. Testing for ```UBLK_U_CMD_START_DEV``` is fine, ```CTRL_CMD_HAS_DATA``` is not. And ```struct ublk_ctrl_cmd_data``` is not a public struct. There's no way the ublk tests could ever get enabled. I'll fix it up.

Issue #1014, opened 2023-12-13
Question regarding behavior of IORING_POLL_ADD_MULTI


man9ourah:
Hello,

my kernel version: 6.2.0-32-generic

is IORING_POLL_ADD_MULTI supposed to be edge-triggered or level-triggered? Right now I am using it and assuming that it is the former and expecting a cqe for when an event is set and then when it is cleared. For example, if I got a poll cqe with POLLIN event and handled it fully with read, will I get another poll cqe stating the new status of the event? 

Just to share what I really want: I want my application to be up-to-date with the status of the event on a fd. Whenever it changes to either state, I want to receive a cqe with the new state. Is that possible?

Just adding references here to why I thought it is edge-triggered: 

From: https://www.spinics.net/lists/io-uring/msg10211.html
> IORING_POLL_ADD_MULTI is similar to epoll's edge-triggered mode,

Also the new (maybe still not yet active) IORING_POLL_ADD_LEVEL flag means currently it is edge? 

Could you please help me understand? I think maybe I have things backward? 

Thanks!
Mansour.

sxstd001:
I tested it, and the effects are the same.
**io_uring_prep_poll_add == io_uring_prep_poll_multishot == epoll Level Triggered;**

isilence:
In short, they're rather level-triggered.

`io_uring_prep_poll_add` is purely level-triggered. Similarly, when you just queued a `io_uring_prep_poll_multishot` it's also level-triggered, i.e. if there's data/event/etc. at the time you submit the request it'll post a completion. However, for all subsequent completions of a multishot poll it kind of resembles the edge mode because it'll be posting completions only in reaction to new events/data arriving.

Issue #990, opened 2023-11-16
Ability to pass file range to IORING_OP_FSYNC is undocumented


RedBeard0531:
While looking through the implementation in the kernel, I noticed that the code [takes the `off` and `len` flags from the sqe](https://github.com/torvalds/linux/blob/c42d9eeef8e5ba9292eda36fd8e3c11f35ee065c/io_uring/sync.c#L64-L65), and [passes them to `vfs_fsync_range`](https://github.com/torvalds/linux/blob/c42d9eeef8e5ba9292eda36fd8e3c11f35ee065c/io_uring/sync.c#L73-L80). I'm not an expert in the linux kernel, but based on the fact that it is still calling `vfs_fsync_range` rather than `sync_file_range`, so I'm hopeful the it avoids the [pitfall](https://man.archlinux.org/man/sync_file_range.2.en#Warning) that makes the latter completely unusable for durability. This is however completely undocumented, so I'd be a bit nervous relying on it. Was it a documentation oversight, or was it intentionally undocumented because it shouldn't be used? If intentional, should the code be removed from the kernel? If an oversight, should it also be exposed from liburing?

For context, I'm interested in this for forcing durability in an append-only* write-ahead log. The application is constantly pushing more data onto the file using `IORING_OP_WRITEV`, and also tries to always have a `IORING_OP_FSYNC` in flight (as long as there are unsynced writes). As soon as the CQE for one fsync comes in, I submit another one with `IO_SQE_DRAIN` and remember the ending offset of the last submitted write. Because I am holding up notification of durability for all ops written before that offset until the CQE comes back, I want it to come back as quickly as possible, and ideally not wait for any writes past that offset (which will still be submitted while the fsync is in flight), since they won't be able to be acknowledged anyway until the next fsync completes. Is this a valid use case for setting `off` and `len` on the SQE?

\* Technically it is append-only at the logical level, but not at the FS layer. We preallocate the file by writing zero blocks before storing any actual data because that seems to be the only way to get fsyncs to complete with a latency of 1 disk write rather than 2 (at least on xfs and ext4). Even using fallocate still needs to do metadata writes.

axboe:
It should be supported like that, it's just the documentation that is lacking. If you are so inclined, please submit a man page update for it!

PR #1225, opened 2024-09-07
[GIT PULL] man/io_uring_enter: Add missing Op codes and some minor other additions

CPestka:
This adds documentation for the following missing Op codes:

1. IORING_OP_GETXATTR
2. IORING_OP_SETXATTR
3. IORING_OP_FGETXATTR
4. IORING_OP_FSETXATTR
5. IORING_OP_BIND
6. IORING_OP_LISTEN
7. IORING_OP_FTRUNCATE
8. IORING_OP_SENDMSG_ZC
9. IORING_OP_READ_MULTISHOT
10. IORING_OP_FUTEX_WAIT
11. IORING_OP_FUTEX_WAITV
12. IORING_OP_FUTEX_WAKE

It also adds documentation about:
1. the undocumented ability to sync file ranges rather than the whole file with IORING_OP_FSYNC see #990
2. the "level" and "edge triggered" behavior of IORING_OP_POLL_ADD see #1014
3. how to enforce ordering between requests after the remark in IORING_OP_FSYNC how one can not rely on ordering per default

As it seems there are no man pages for the new futex syscalls yet (they are normally done by glibc right?) I'm not very sure that I got everything right regarding these syscalls.

Two Ops that are still missing now are IORING_OP_FIXED_FD_INSTALL and IORING_OP_URING_CMD.
IORING_OP_URING_CMD should maybe have its own man page? Currently it is only nvme that is using it, but i supposed the description of it could grow large over time. (As a side note liburing helper functions for nvme cmds would be neat :) )
Regarding IORING_OP_FIXED_FD_INSTALL I wasn't entirely sure how it works. It is apparently used to take "ownership" (installing it in the reg file table and dealing with the file refcount)  of a registered file of another ring, but I wasn't sure how the passing of the file struct is actually happening. Does that have smth to do with MSG rings?

On another side note:
IORING_OP_FIXED_FD_INSTALL, IORING_OP_FTRUNCATE, IORING_OP_BIND, IORING_OP_LISTEN seem to be missing in the enum in tools/include/uapi/linux/io_uring.h, right? 

----
## git request-pull output:
```
The following changes since commit 0fe5c09195c0918f89582dd6ff098a58a0bdf62a:

  configure: fix ublk_cmd header check (2024-09-06 15:54:04 -0600)

  https://github.com/CPestka/liburing io_uring_enter_man

for you to fetch changes up to c6a303d012bf453cf2dc5f593ed80814715d511b:

  man/io_uring_enter: Adds docs about the capability to fsync file ranges (2024-09-07 16:28:49 +0200)

CPestka (11):
      man/io_uring_enter: Add Docs for missing xattr related OP codes
      man/io_uring_enter: Add docs for io_uring Op code for bind(2)
      man/io_uring_enter: Adds doc for io_uring OP code of listen(2)
      man/io_uring_enter: Adds docs for io_uring op for ftruncate(2)
      man/io_uring_enter: Add docs for zero copy version of sendmsg op
      man/io_uring_enter: Add docs for multishoot variant of read OP
      man/io_uring_enter: Add doc for io_uring version of futex_wait(2)
      man/io_uring_enter: Adds docs for io_uring version of futex_wake(2)
      man/io_uring_enter: Add docs for the io_uring version of futex_waitv(2)
      man/io_uring_enter: Added remark about ordering of SQE/CQEs
      man/io_uring_enter: Adds docs about the capability to fsync file ranges

 man/io_uring_enter.2 | 182 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 1 file changed, 182 insertions(+)
```
----
<details>

## Pull Request Guidelines
```
```
13. Follow the commit message format rules below.
14. Follow the Linux kernel coding style (see: https://github.com/torvalds/linux/blob/master/Documentation/process/coding-style.rst).

5. Then an empty line.
6. Then a description (may be omitted for truly trivial changes).
15. Then an empty line again (if it has a description).
16. Then a `Signed-off-by` tag with your real name and email. For example:
```
```

certain specific format).

of truth, not github.

URL.

```
```

Use a real email address!

### Commit message example:
```

safe without synchronization.

```

</details>

----

### DIFF
--- a/man/io_uring_enter.2
+++ b/man/io_uring_enter.2
@@ -1258,6 +1258,22 @@ for details on how to setup a context for fixed buffer I/O.
 .PP
 
 .TP
+.B IORING_OP_SENDMSG_ZC
+Issue the zerocopy equivalent of a
+.BR sendmsg (2)
+system call.
+Works just like
+.IR IORING_OP_SENDMSG ,
+but like
+.I IORING_OP_SEND_ZC
+supports
+.IR IORING_RECVSEND_FIXED_BUF .
+For additional notes regarding zero copy see
+.IR IORING_OP_SEND_ZC .
+
+Available since 6.1
+
+.TP
 .B IORING_OP_WAITID
 Issue the equivalent of a
 .BR waitid(2)
--- a/man/io_uring_enter.2
+++ b/man/io_uring_enter.2
@@ -1370,6 +1370,23 @@ must contain the length to which the file will be truncated.
 
 Available since 6.9.
 
+.TP
+.B IORING_OP_READ_MULTISHOT
+Like
+.IR IORING_OP_READ ,
+but similar to requests prepared with
+.IR io_uring_prep_multishoot_accept (3)
+additional reads and thus CQEs will be performed based on this single SQE once
+there is more data available.
+Is restricted to pollable files and will fall back to single shoot if the file
+does not support
+.IR NOWAIT .
+Like other multishot type requests, the application should look at the CQE
+flags and see if
+.I IORING_CQE_F_MORE
+is set on completion as an indication of whether or not the read request will
+generate further CQEs.
+
 .PP
 The
 .I flags
--- a/man/io_uring_enter.2
+++ b/man/io_uring_enter.2
@@ -1329,31 +1329,6 @@ must contain the length of the address.
 
 Available since 6.11.
 
-.TP
-.B IORING_OP_LISTEN
-Issues the equivalent of the
-.IR listen(2)
-system call.
-.I fd
-must contain the file descriptor of the socket and
-.I addr
-must contain the backlog parameter, i.e. the maximum amount of pending
-queued connections.
-
-Available since 6.11.
-
-.TP
-.B IORING_OP_FTRUNCATE
-Issues the equivalent of the
-.IR ftruncate(2)
-system call.
-.I fd
-must contain the file descriptor of the file to truncate and
-.I off
-must contain the length to which to file will be truncated.
-
-Available since 6.9.
-
 .PP
 The
 .I flags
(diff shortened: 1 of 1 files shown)


axboe:
Thanks for doing this!

> IORING_OP_URING_CMD should maybe have its own man page? Currently it is only nvme that is using it

I think it should have its own man page, but it's not true that only nvme is using it. Yes it's used for nvme passthrough, but if you look at the socket op methods for get/setsockopt, then those use the uring_cmd interface as well. And that part will only grow. You can consider OP_URING_CMD as a way of doing per-file type private operations, kind of like ioctls (except async).

> IORING_OP_URING_CMD should maybe have its own man page? Currently it is only nvme that is using it

It takes a registered file index and installs it in the normal process file table, turning it into a file descriptor you can use with regular syscalls as well. It remains a registered descriptor as well.

CPestka:
> It takes a registered file index and installs it in the normal process file table, turning it into a file descriptor you can use with regular syscalls as well. It remains a registered descriptor as well.

Oh, that makes a lot of sense. I was wondering a while back how one would e.g. call smth like getpeername() on a connection made with accept using registered fds. I guess that's the answer ^^

CPestka:
Btw it seems like the ability to pass the new "non-registered" fd back to the user is implemented in receive_fd(), but not wired up to the IORING_OP_FIXED_FD_INSTALL, which always passes it NULL for the user space ptr. Is that an oversight or intentional?

axboe:
> Btw it seems like the ability to pass the new "non-registered" fd back to the user is implemented in receive_fd(), but not wired up to the IORING_OP_FIXED_FD_INSTALL, which always passes it NULL for the user space ptr. Is that an oversight or intentional?

That's intentional - the userptr doesn't exist on the io_uring side. If you look at receive_fd(), it both returns the fd value where the regular file descriptor was installed, and copies it to a userspace address, if one is given. io_uring just uses the former.

PR #1236, opened 2024-09-19
test/linked-defer-close: make test work without improper IOSQE_IO_LINK

Cydox:
This makes the linked-defer-close test work correctly without relying on
improper usage of IOSQE_IO_LINK.

The test will now issue three sends with
IOSQE_CQE_SKIP_SUCCESS | IOSQE_IO_LINK followed by a close with
IOSQE_CQE_SKIP_SUCCESS on a ring setup with IORING_SETUP_DEFER_TASKRUN.

On an incorrectly behaving kernel the connection will not be closed and
remain open.

This test first fails on kernel commit 846072f16eed3b3fb4e59b677f3ed8afb8509b89

Signed-off-by: Jan Hendrik Farr <kernel@jfarr.cc>

### DIFF
--- a/test/linked-defer-close.c
+++ b/test/linked-defer-close.c
@@ -25,6 +25,7 @@ enum {
 	IS_ACCEPT = 0,
 	IS_SEND = 0x100,
 	IS_SEND2 = 0x101,
+	IS_SEND3 = 0x102,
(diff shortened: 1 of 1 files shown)


axboe:
Thanks!

PR #1244, opened 2024-09-24
[GIT PULL] Fix some syscall name typos in man pages

aconz2:
The man pages reference `unlinkat2`, `mkdirat2`, `symlinkat2`, and `linkat2` which I think are typos. This fixes them by removing the `2`.

Related to https://github.com/tokio-rs/io-uring/pull/304

There are also a few references to `futex2(2)` which I now understand (I think) refers to the group `futex_{wait,waitv,wake}`, but unfortunately none of those have man pages. Is there a place that documents those that the `futex2(2)` link could reference instead?

----
## git request-pull output:
```
The following changes since commit 0c8975ad7c95de0f81a9dc869b55e77131159040:

  man/io_uring_register_ring_fd: add note about kernel availability (2024-09-22 00:47:34 -0600)

  https://github.com/aconz2/liburing fix-man-pages2

for you to fetch changes up to d9716086048112ad95a1e976005ed70ddc14f8e5:

  Fix some syscall name typos in man pages (2024-09-24 09:00:14 -0500)

Andrew Consroe (1):
      Fix some syscall name typos in man pages

 man/io_uring_enter.2 | 8 ++++----
 1 file changed, 4 insertions(+), 4 deletions(-)
```
----
<details>

## Pull Request Guidelines
```
```

2. Then an empty line.
```
```

certain specific format).

of truth, not github.

URL.

```
```

Use a real email address!

### Commit message example:
```

safe without synchronization.

```

</details>

----

### DIFF
--- a/man/io_uring_enter.2
+++ b/man/io_uring_enter.2
@@ -1135,7 +1135,7 @@ Available since 5.11.
 .TP
 .B IORING_OP_UNLINKAT
 Issue the equivalent of a
-.BR unlinkat2 (2)
+.BR unlinkat (2)
 system call.
 .I fd
 should be set to the
@@ -1154,7 +1154,7 @@ Available since 5.11.
 .TP
 .B IORING_OP_MKDIRAT
 Issue the equivalent of a
-.BR mkdirat2 (2)
+.BR mkdirat (2)
 system call.
 .I fd
 should be set to the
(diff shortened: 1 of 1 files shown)


axboe:
There are no futex2 man pages yet it seems, but they should get there at some point... I don't think there are better resources we can point to right now, so probably just leave those as-is? The kernel headers are probably the only reference for them a this point.

PR #1256, opened 2024-10-05
man/io_uring_internal: Man page about high lvl inner workings of io_uring

CPestka:
This adds the first of the new man 7 pages suggested in #1241

It contains a high lvl overview of what needs to be done on the kernel side for all requests, who does the work with a given configuration, explains the async punt and describes what io_uring kernel threads exist in which case.

Let me know if I got anything wrong, smth is missing etc..
Also, wasn't sure about the name of the page.

----
## git request-pull output:
```
The following changes since commit 206650ff72b6ea4d76921f9c91ebfffd9902e6a0:

  test/fixed-hugepage: skip test on -ENOMEM (2024-09-27 10:27:10 -0600)

  https://github.com/CPestka/liburing man_internals

for you to fetch changes up to fc5266b9626ac9798dce455f031871f4817c7cea:

  man/io_uring_internal: Add man page about relevant internals for users (2024-10-05 15:00:14 +0200)

CPestka (1):
      man/io_uring_internal: Add man page about relevant internals for users

 man/io_uring_internals.7 | 225 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 1 file changed, 225 insertions(+)
 create mode 100644 man/io_uring_internals.7
```
----
<details>

## Pull Request Guidelines
```
```

2. Then an empty line.
```
```

certain specific format).

of truth, not github.

URL.

```
```

Use a real email address!

### Commit message example:
```

safe without synchronization.

```

</details>

----


axboe:
Thanks for kicking this off! I'll add some comments in the diff.

man/io_uring_internals.7
```diff
@@ -0,0 +1,225 @@
...
+.SH SYNOPSIS
+.nf
+.B "#include <linux/io_uring.h>"
+.fi
+.PP
+.SH DESCRIPTION
+.PP
+.B io_uring
+is a linux specific, asynchronous API that allows the submission of requests to
+the kernel that are typically otherwise performed via a syscall. Requests are
```
axboe, line 13: Typically yes, but not exclusively. Not sure if it bears mentioning or not...

man/io_uring_internals.7
```diff
@@ -0,0 +1,225 @@
...
+passed to the kernel via a shared ring buffer the
+.I Submission Queue
+(SQ) and completion notifications are passed back to the application via the
+.I Completion Queue
+(CQ). An important detail here is that after a request has been submitted to
```
axboe, line 18: Not sure that sentence reads that well, I'm having a hard time trying to make sense of it.

man/io_uring_internals.7
```diff
@@ -0,0 +1,225 @@
...
+thread for the threaded IRQ handling gets scheduled, if that is used.
+
+.PP
+.SH The Submission Side Work
+.PP
+
+The tasks required in kernel space on the submission side are mostly checking
+the SQ for newly arrived SQEs, parsing and check them for validity and
+permissions and then passing them on to the responsible system, such as a
+block device driver. An important note here is that
```
axboe, line 47: I'd mention "such as a block device driver, networking stack, etc" or something like that. Don't want to make this sound storage centric.

man/io_uring_internals.7
```diff
@@ -0,0 +1,225 @@
...
+syscall made to submit the new requests, will never block. However,
+.I io_uring
+relies on the capabilities of the responsible system to perform the submission
+without blocking.
+.I io_uring
+will first attempt to submit the request without blocking.
+If this fails, e.g. due to the respective system not supporting non-blocking
+submissions,
+.I io_uring
+will
```
axboe, line 61: This isn't fully accurate. If an IO can be issued in a non-blocking fashion, then one of two things can happen:

1) It's done. Examples of this would be writing to a pipe/socket (for example), reading from a pipe/socket, or reading/writing to/from a regular file where the result is either in the page cache already (for a read), or io_uring was able to just copy it to the page cache (for a write). For these cases, a CQE will be posted immediately, even before io_uring_enter(2) returns.

2a) It wasn't done, but submitted async. io_uring will get a callback at some point when the operation completes, and a CQE will be posted. Examples of this are async reads/writes to a storage device.

2b) It wasn't done, but the file in question can signal readiness for when the operation can be retried. Examples of this are any pollable file, like a pipe, socket, etc. When io_uring receives the callback that data can now be read/written, it will retry the operation. Importantly, this retry happens from the task that submitted the IO. There's no async thread involved in this operation.

2c) It wasn't done, and the file has limited async support. Eg it cannot signal when it's ready to do IO. For this case, and only this case, does io_uring punt to an async worker to do the IO.

I don't want to imply that io_uring just willy nilly punts to async workers, as that is not the case, and that would not be very efficient. It's a last resort kind of thing, for when the driver / file type is pretty basic and doesn't support more than very basic primitives.

Now, for the application, it doesn't really matter which of the 2 cases end up happening, as completions are posted as it expects. But for efficiency reasons, it very much does matter, and there's a common theme where people assume that io_uring is just a thread work pool. That is very much WRONG, and this man page should not perpetuate that myth, it should help clear up the misunderstanding.

man/io_uring_internals.7
```diff
@@ -0,0 +1,225 @@
...
+.I async punt
+the request, i.e. off-load these requests to the
+.I IO work queue
+(IO WQ) (see description below). 
+
+.PP
+.SH The Completion Side Work
+.PP
+
+The tasks required in kernel space on the completion side mostly come in the
```
axboe, line 71: General theme - task is not the best word to use, because it implies a relationship to a thread/process. Not sure what's a better word to use here, just tossing it out there.

man/io_uring_internals.7
```diff
@@ -0,0 +1,225 @@
...
+.SH Who does the work
+.PP
+
+One of
+the primary motivations behind
+.I io_uring
+was to reduce or entirely avoid the overheads of syscalls to provide the
+required CPU time in kernel space. The mechanism that
+.I io_uring
+utilizes to achieve this differs depending on the configuration with different
```
axboe, line 86: utilize

man/io_uring_internals.7
```diff
@@ -0,0 +1,225 @@
...
+trade-offs between configurations in respect to e.g. CPU efficiency and latency.
+
+With the default configuration the primary mechanism to provide the kernel space
```
axboe, line 89: Not sure what this "provide the kernel space CPU time" means here?

man/io_uring_internals.7
```diff
@@ -0,0 +1,225 @@
...
+requests for submissions and process arrived completions within the same
+.IR io_uring_enter (2)
+call. Applications can set the flag
+.I IORING_ENTER_GETEVENTS
+to in addition to processing any pending submissions, process any arrived
+completions and
+optionally wait until a specified amount of completions have arrived before
+returning.
+
+If polled I/O is used all completion related work is performed during the
```
axboe, line 115: I like this section!

man/io_uring_internals.7
```diff
@@ -0,0 +1,225 @@
...
+.I IORING_ENTER_GETEVENTS
+set, rather than at any context switch, which gives the application more agency
+about when the work is executed, thus enabling e.g. more opportunities for
+batching.
+
+.PP
+.SH Submission Queue Polling
+.PP
+
+Sq polling introduces a dedicated kernel thread that performs essentially all
```
axboe, line 155: Here, and in other spots, it's important to note that io_uring does NOT utilize any kernel threads. In Linux, a kernel thread is a special kind of thread that is entirely decoupled from any other real process/thread running in userspace. It doesn't have any files, mm, etc associated with it.

What io_uring uses are "io threads", which are exactly like a thread created with eg pthread_create() in the sense that they share any resources that the original task has, and any credentials, namespaces, etc. The only thing that makes them different is that they are created by io_uring, and they never exit to userspace. They sit around and do work, if needed, and then go away when they are no longer needed.

Hence I would probably explain this io thread concept when it's initially encountered in this man page, and then subsequently refer to io threads and remove any mention of kernel threads.

man/io_uring_internals.7
```diff
@@ -0,0 +1,225 @@
...
+submission and completion related tasks from fetching SQEs from the SQ,
+submitting requests, polling requests, if configured for I/O poll and posting
+CQEs. Notably, async punt requests are still processed by the IO WQ, to not
```
axboe, line 158: See above explanation of how requests are issued and when io-wq is actually used, applies here too.

man/io_uring_internals.7
```diff
@@ -0,0 +1,225 @@
...
+to do for a user supplied timeout it goes to sleep. Sq polling removes the need
+for any syscall during operation, besides waking up the sq thread after long
+periods of inactivity and thus reduces per request overheads at the cost of a
+high constant upkeep cost.
+
+.PP
+.SH IO Work Queue
+.PP
+
+The IO WQ is a kernel thread pool used to execute any requests that can not be
```
axboe, line 169: io-wq is a pool of io threads [...]

man/io_uring_internals.7
```diff
@@ -0,0 +1,225 @@
...
+submitted in a non-blocking way to the underlying subsystem, due to missing
+support in said subsystem. After either the sq poll thread or a user space
+thread calling
+.IR io_uring_enter (2)
+fails the initial attempt to submit the request without blocking it passes the
+request on to a IO WQ thread that then performs the blocking submission. While
+this mechanism ensures that
+.IR io_uring ,
+unlike e.g. AIO, never blocks on any of the submission paths, it is, as the
```
axboe, line 178: in any of the submission paths.

Stop the sentence there. And then I don't understand what the rest of that original sentence is trying to say?

man/io_uring_internals.7
```diff
@@ -0,0 +1,225 @@
...
+name of this mechanism, the async punt, suggests not ideal. The blocking
+nature of the submission, the passing of the request to another thread, as
+well as the scheduling of the IO WQ threads are all ideally avoided
+overheads. Significant IO WQ activity can thus be seen as an indicator that
```
axboe, line 182: This is very true.

man/io_uring_internals.7
CPestka: Yeah, will do.

man/io_uring_internals.7
CPestka: I forgot that 1. is even a thing when writing this :D, so this should definitively be mentioned explicitly.
Also explaining 2a) and 2b) in more detail like this is probably a good idea.

man/io_uring_internals.7
CPestka: Yeah true. Many of the alternatives are also overloaded. I guess "work" would be better. The heading already uses it.

man/io_uring_internals.7
CPestka: I think utilizes is correct here?

man/io_uring_internals.7
CPestka: "Someone" needs to run some code in kernel space (to do the kernel side submission, post the CQE etc.) , be that one of the submitting processes after the context switch during a syscall or e.g. the sq poll thread or to a limited extend the io wq threads. So "Someone" e.g. the caller of io_uring_enter or the sq poll thread would "provide the kernel space CPU time" ... and use it to run the relevant code in kernel space. That's how i have been thinking about this, but yeah maybe not the best wording...

man/io_uring_internals.7
CPestka: Ah I see, my mistake. I thought: thread that never exits to user space i.e. only runs in kernel space == kernel thread. The distinction makes a lot of sense though. I assume this is done for permission management on files (credentials) and zero copy stuff (mm)? I only noticed the difference till now, by io_urings worker threads showing up in traces with tracy or perf, where this did not seem to be the case for other (or what i thought were) other kernel threads.

man/io_uring_internals.7
CPestka: y. Cutting the sentence there makes sense. What I tried to say there is mostly redundant with the following sentences anyways. That being that one could have probably guessed that its not ideal only by the name. It is called the async *punt* for a reason. It's a fall back. There is probably a reason why it is not the first thing that is attempted. Anyways, not really a reason why that comment should be here.

CPestka:
@axboe  Thanks for the notes! I left some comments and will type up the corrections tomorrow if i get to it.

CPestka:
Ok, finally got around to address your comments @axboe. The main things, I guess, are: I added a small section for explaining the io threads separately and expanded the explanation for scenarios during submission. Also a couple of other minor rewordings etc. Let me know if I missed smth or got something wrong in the corrections :)

axboe:
Thanks, I'll take another look!

man/io_uring_internals.7
```diff
@@ -0,0 +1,282 @@
...
+as different syscalls types can be freely mixed and matched and chains of
+dependent requests, such as a
+.IR send (2)
+followed by a
+.IR recv (2)
+can be submitted with one syscall. Furthermore it is possible to both process
+requests for submissions and process arrived completions within the same
+.IR io_uring_enter (2)
+call. Applications can set the flag
+.I IORING_ENTER_GETEVENTS
```
redbaron, line 139: for someone who doesn't have deep understanding of io_uring (me) it is still not clear what `IORING_ENTER_GETEVENTS` actually do. Wording says that it allows processing completions in `io_uring_enter` , how do completions processed when this flag is missing then?

zea64: I believe it just makes io_uring_enter wait for completions before returning. The wording is quite awkward here, I'd suggest something like "Applications can set the flag IORING_ENTER_GETEVENTS to process submissions and then wait for completions (or a timeout) in one syscall."

PR #1328, opened 2025-01-15
man/io_uring_prep_recv: expand on how to handle bundles

romange:
Specifically, explain how to identify multiple buffer ids upon bundle completion.

----
## git request-pull output:
```
The following changes since commit a5dd7f85d4b667aeb687afc562ff9750044b42c2:

  man/io_uring_clone_buffers: mention that both rings must share mm (2025-01-14 11:06:56 -0700)

  https://github.com/romange/liburing Pr1

for you to fetch changes up to 73bb0ac90437f87b8ecbde50fe30fa25a4df1ae5:

  man/io_uring_prep_recv: expand on how to handle bundles (2025-02-03 14:37:41 +0200)

Roman Gershman (1):
      man/io_uring_prep_recv: expand on how to handle bundles

 man/io_uring_prep_recv.3 | 14 ++++++++++----
 1 file changed, 10 insertions(+), 4 deletions(-)
```
----
<details>

## Pull Request Guidelines
```
```

2. Then an empty line.
```
```

certain specific format).

of truth, not github.

URL.

```
```

Use a real email address!

### Commit message example:
```

safe without synchronization.

```

</details>

----

### DIFF
--- a/man/io_uring_prep_recv.3
+++ b/man/io_uring_prep_recv.3
@@ -94,12 +94,18 @@ field indicates the total number of bytes received, and the buffer ID returned
 in the CQE
 .I flags
 field indicates the first buffer in the receive operation. The application must
-iterate from the indicated initial buffer ID and until all
+process the indicated initial buffer ID and until all
 .I res
 bytes have been seen to know which is the last buffer in the receive operation.
-The buffer IDs consumed will be contiguous from the starting ID, in the order
-in which they were added to the buffer ring used. Receiving in bundles can
-improve performance when more than one chunk of data is available to receive,
+The buffers consumed will be contiguous from the initial buffer, in the order
+in which they appear in the buffer ring. The CQE struct does not contain
+the position of the buffer in the buffer ring, therefore in order to identify
+buffers contained by the bundle, it is advised to maintain the cached head
+index per buffer ring. This uint16_t index represents the position of the next
+buffer to be consumed within the ring. Upon completion of a receive operation,
+the cached head index should be incremented accordingly.
+Receiving in bundles can improve performance when more than one chunk of
+data is available to receive,
 by eliminating redundant round trips through the networking stack. Receive
 bundles may be used by both single shot and multishot receive operations. Note
 that, internally, bundles rely on the networking stack passing back how much


romange:
Consider this a  draft. Once the wording is finalized I will copy it to another man page where we duplicate the explanation about bundles. In addition, I can add the necessary code to proxy.c (as part of this PR or separately)

man/io_uring_prep_recv.3
```diff
@@ -97,8 +97,11 @@ field indicates the first buffer in the receive operation. The application must
 iterate from the indicated initial buffer ID and until all
 .I res
 bytes have been seen to know which is the last buffer in the receive operation.
-The buffer IDs consumed will be contiguous from the starting ID, in the order
-in which they were added to the buffer ring used. Receiving in bundles can
+The buffers consumed will be contiguous from the initial buffer, in the order
+in which they appear in the buffer ring. The CQE struct does not contain the position
+of the buffer in the bufring, therefore in order to identify buffers contained by the bundle,
+it is advised to maintain the cached head index per bufring. This uint16_t index represents the position of the next buffer to be consumed within the ring. Upon completion of a receive operation on a buffer ring, the cached head index should be incremented accordingly.
```
axboe, line 103: Please break lines at 80 chars. And use "buffer ring" rather than "bufring", generally that's the term used in man pages.

Outside of that, looks fine to me.

romange:
Done. Also added the proxy changes. How do you usually test proxy? with what loadtest program?

axboe:
> How do you usually test proxy? with what loadtest program?

Most testing I've done has just been with a simple test program I wrote, which spawns N threads sending M sized buffers. In other words, I think anything really will work, just ensure you have recv bundles enabled of course. If it were me, I'd test one in sink mode, and one in fwd mode where it just dumps it to ncat or something.

romange:
@axboe  I've done some basic testing with proxy but i am not really happy with the level of testing I could do. 
When I tried to do something more sophisticated I've stumbled upon #1338 (not related to my changes) which prevents me to test it further. Unfortunately, I do not have time to fix #1338, so there are two options here:
1. I am reverting my proxy changes and submit only the docs PR.
2. You are ok with submitting my proxy changes based on your review.

axboe:
Let's do the man page separately, it really should be two separate commits to begin with as they aren't directly related. A separate PR for the proxy change would be useful, then I can test it (and fix that other issue) when time arises. It's not really that urgent.

PR #1336, opened 2025-01-27
[GIT PULL] Hide io_uring_get_sqe symbol

cmazakas:
The user-facing API for liburing has the user interface with io_uring_get_sqe via a static inline function. However, this function is also defined in the plain liburing archive, which can potential lead to ODR violations.

We instead define it properly only for the liburing-ffi case and keep it hidden when the liburing archive itself is built.

See:
```
exbigboss@pleiades ~/cpp/liburing (symbol-fixes)
❯ nm -an src/liburing-ffi.a | grep io_uring_get_sqe
00000000000013d0 T _io_uring_get_sqe
0000000000001540 T io_uring_get_sqe
exbigboss@pleiades ~/cpp/liburing (symbol-fixes)
❯ nm -an src/liburing.a | grep io_uring_get_sqe
exbigboss@pleiades ~/cpp/liburing (symbol-fixes) [1]
❯
```

----
## git request-pull output:
```
The following changes since commit 6c509e2b0c881a13b83b259a221bf15fc9b3f681:

  test/sqwait: cleanup allocated iovecs (2025-01-22 19:48:36 -0700)

  https://github.com/cmazakas/liburing.git symbol-fixes

for you to fetch changes up to 659c39092a7ddfaa68e4f4560475a0e9232af361:

  Hide io_uring_get_sqe symbol (2025-01-27 09:22:34 -0800)

Christian Mazakas (1):
      Hide io_uring_get_sqe symbol

 debian/liburing2.symbols | 1 -
 src/ffi.c                | 5 +++++
 src/include/liburing.h   | 2 --
 src/liburing.map         | 1 -
 src/queue.c              | 4 +++-
 5 files changed, 8 insertions(+), 5 deletions(-)
```

----

### DIFF
--- a/debian/liburing2.symbols
+++ b/debian/liburing2.symbols
@@ -11,7 +11,6 @@ liburing.so.2 liburing2 #MINVER#                                                
  io_uring_get_events@LIBURING_2.3 0.7-1
  io_uring_get_probe@LIBURING_2.0 0.7-1
  io_uring_get_probe_ring@LIBURING_2.0 0.7-1
- io_uring_get_sqe@LIBURING_2.0 0.7-1
  io_uring_mlock_size@LIBURING_2.1 0.7-1
  io_uring_mlock_size_params@LIBURING_2.1 0.7-1
  io_uring_peek_batch_cqe@LIBURING_2.0 0.7-1
--- a/src/ffi.c
+++ b/src/ffi.c
@@ -10,6 +10,11 @@
 
 #include "liburing.h"
 
+struct io_uring_sqe *io_uring_get_sqe(struct io_uring *ring)
+{
+	return _io_uring_get_sqe(ring);
+}
+
 #ifdef __clang__
 #pragma clang diagnostic pop
 #endif
--- a/src/include/liburing.h
+++ b/src/include/liburing.h
@@ -1624,8 +1624,6 @@ IOURINGINLINE struct io_uring_sqe *io_uring_get_sqe(struct io_uring *ring)
 {
 	return _io_uring_get_sqe(ring);
 }
-#else
-struct io_uring_sqe *io_uring_get_sqe(struct io_uring *ring);
 #endif
 
 ssize_t io_uring_mlock_size(unsigned entries, unsigned flags);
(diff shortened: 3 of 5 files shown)


ammarfaizi2:
No, you can't omit that.

It will break applications that are linked against liburing-2.0 and
earlier versions. Jens did it on purpose to maintain backward
compatibility after marking `io_uring_get_sqe` as a static inline
function.

```
From: Jens Axboe <axboe@kernel.dk>
Date: Wed, 23 Feb 2022 16:28:09 -0700
Subject: [PATCH] queue: provide io_uring_get_sqe() symbol again

With a recent change, io_uring_get_sqe() was moved to the header file
to avoid a function call for this hot function. However, this will break
updates of the library, where the symbol is then not provided:

./app: symbol lookup error: ./app: undefined symbol: io_uring_get_sqe, version LIBURING_2.0

Provide one internally so we still export it, but allow newly compiled
applications to grab the one from the header. That should provide the
best of both worlds.

Fixes: 6cf23f9a6470 ("Move io_uring_get_sqe() inline in the liburing.h header")
---
 src/Makefile           |  2 +-
 src/include/liburing.h |  4 ++++
 src/queue.c            | 14 ++++++++++++++
 3 files changed, 19 insertions(+), 1 deletion(-)
```

ammarfaizi2:
> However, this function is also defined in the plain liburing archive, which can potential lead to ODR violations.

Do you have a real example of an ODR violation that leads to catastrophic
consequences? (e.g., compiler error, runtime crash, etc.).

cmazakas:
> It will break applications that are linked against liburing-2.0 and
> earlier versions. Jens did it on purpose to maintain backward
> compatibility after marking `io_uring_get_sqe` as a static inline
> function.

Ah, this makes a ton of sense, actually.

I wasn't aware of the history of the change. I thought it was an honest mistake. I hadn't considered it's a break for existing applications which rely on semver.

> Do you have a real example of an ODR violation that leads to catastrophic
> consequences? (e.g., compiler error, runtime crash, etc.).

Nope. The concern here is mostly theoretical.

The extra symbol was just something I ran into during development with liburing locally and I figured I'd make a PR to "fix the bug".

Thanks for the explanation. There's nothing else to do for now so I'll just simply close the PR.

PR #1351, opened 2025-02-17
test: Including missing string.h for memcpy

kraj:


ammarfaizi2 (approved):
That's surprising...

How did we miss this?

It has never failed in GitHub CI, and no one has reported build errors.

It seems like a header must have been indirectly including `<string.h>`, but only under certain OS or environments.

kraj:
> That's surprising...
> 
> How did we miss this?
> 
> It has never failed in GitHub CI, and no one has reported build errors.
> 
> It seems like a header must have been indirectly including `<string.h>`, but only under certain OS or environments.

Yes. This is found with latest musl based systems. Musl tends to find the indirect includes issues which we do not see usually with glibc. This started to happen with faily recent musl

PR #1400, opened 2025-05-14
[GIT PULL] register: use io_uring_rsrc_update instead of io_uring_files_update

calebsander:
`struct io_uring_files_update` has a comment saying it is deprecated in favor of `struct io_uring_rsrc_update`. So switch the 2 references to it to use `io_uring_rsrc_update` instead.

----
## git request-pull output:
```
The following changes since commit 47eacd321f107eb53b1d872d5a376ba7885e8e76:

  Merge branch 'master' of https://github.com/Sberm/liburing (2025-05-07 17:03:21 -0600)

  git@github.com:calebsander/liburing.git fix/io_uring_rsrc_update

for you to fetch changes up to 16777f99dc780b1e49407dcff63d64e137c3ff37:

  register: use io_uring_rsrc_update instead of io_uring_files_update (2025-05-09 18:53:27 -0600)

Caleb Sander Mateos (1):
      register: use io_uring_rsrc_update instead of io_uring_files_update

 man/io_uring_register.2 | 2 +-
 src/register.c          | 4 ++--
 2 files changed, 3 insertions(+), 3 deletions(-)
```
----

### DIFF
--- a/man/io_uring_register.2
+++ b/man/io_uring_register.2
@@ -304,7 +304,7 @@ or replacing an existing entry with a new existing entry.
 .I arg
 must contain a pointer to a
 .I struct
-.IR io_uring_files_update ,
+.IR io_uring_rsrc_update ,
 which contains
 an offset on which to start the update, and an array of file descriptors to
 use for the update.
--- a/src/register.c
+++ b/src/register.c
@@ -111,9 +111,9 @@ int io_uring_register_files_update(struct io_uring *ring, unsigned off,
 {
 	liburing_sanitize_address(files);
 
-	struct io_uring_files_update up = {
+	struct io_uring_rsrc_update up = {
 		.offset	= off,
-		.fds	= (unsigned long) files,
+		.data	= (unsigned long) files,
 	};
 
 	return do_register(ring, IORING_REGISTER_FILES_UPDATE, &up, nr_files);


isilence:
It still uses `io_uring_files_update` because switching will break liburing with older kernels. We need to check first with which kernel release it came about and if it's still supported.

axboe:
@isilence ah indeed, how did I miss that... liburing is always kernel independent. I'll revert the change for now. Caleb, I'll revert the code part of the change.

calebsander:
> switching will break liburing with older kernels.

How do you mean? The structs have the same layouts. And newer kernels already use `io_uring_rsrc_update` and there don't appear to be any backwards compatibility issues with existing versions of liburing that use `io_uring_files_update`.

axboe:
Indeed, it's the same op still, my memory is bad. So change should be fine, even if a bit misleading.

isilence:
> > switching will break liburing with older kernels.
> 
> How do you mean? The structs have the same layouts. And newer kernels already use `io_uring_rsrc_update` and there don't appear to be any backwards compatibility issues with existing versions of liburing that use `io_uring_files_update`.

Yeah,, I mixed it up with 2nd versions of the rsrc struct.

PR #1436, opened 2025-07-13
Do not discard high 32 bits of length in send

### LINKED ISSUE #1435: `io_uring_prep_send` returning zero in the CQE with polling
I'm witnessing an odd behavior whereby `io_uring_prep_send` SQEs get completed with `res = 0`, even if the connection is still alive and using `send` directly works just fine. This is on 6.10.3 and sending messages down a blocking TCP socket, with a polling uring.

I _think_ this is due to `MSG_DONTWAIT` being passed to `sock_sendmsg`. [`MSG_DONTWAIT` is set when `IO_URING_F_NONBLOCK` is set](https://elixir.bootlin.com/linux/v6.10.3/source/io_uring/net.c#L586), which I think is the case for polling queues. So I'd imagine that when the socket buffer is full you'll just get 0.

Is this expected? I'd expect `send` on blocking sockets to block until space is present.

bitonic:
Fixes #1435

### DIFF
--- a/src/include/liburing.h
+++ b/src/include/liburing.h
@@ -919,17 +919,23 @@ IOURINGINLINE void io_uring_prep_statx(struct io_uring_sqe *sqe, int dfd,
 	sqe->statx_flags = (__u32) flags;
 }
 
+static inline __u32 __io_uring_cap_len(size_t len)
+{
+	size_t mask = -(size_t)(len > UINT32_MAX);
+	return (len & ~mask) | (UINT32_MAX & mask);
+}
+
 IOURINGINLINE void io_uring_prep_fadvise(struct io_uring_sqe *sqe, int fd,
 					 __u64 offset, __u32 len, int advice)
 {
-	io_uring_prep_rw(IORING_OP_FADVISE, sqe, fd, NULL, (__u32) len, offset);
+	io_uring_prep_rw(IORING_OP_FADVISE, sqe, fd, NULL, __io_uring_cap_len(len), offset);
 	sqe->fadvise_advice = (__u32) advice;
 }
 
 IOURINGINLINE void io_uring_prep_madvise(struct io_uring_sqe *sqe, void *addr,
 					 __u32 length, int advice)
 {
-	io_uring_prep_rw(IORING_OP_MADVISE, sqe, -1, addr, (__u32) length, 0);
+	io_uring_prep_rw(IORING_OP_MADVISE, sqe, -1, addr, __io_uring_cap_len(length), 0);
 	sqe->fadvise_advice = (__u32) advice;
 }
 
@@ -951,7 +957,7 @@ IOURINGINLINE void io_uring_prep_madvise64(struct io_uring_sqe *sqe, void *addr,
 IOURINGINLINE void io_uring_prep_send(struct io_uring_sqe *sqe, int sockfd,
 				      const void *buf, size_t len, int flags)
 {
-	io_uring_prep_rw(IORING_OP_SEND, sqe, sockfd, buf, (__u32) len, 0);
+	io_uring_prep_rw(IORING_OP_SEND, sqe, sockfd, buf, __io_uring_cap_len(len), 0);
 	sqe->msg_flags = (__u32) flags;
 }
 
@@ -983,7 +989,7 @@ IOURINGINLINE void io_uring_prep_send_zc(struct io_uring_sqe *sqe, int sockfd,
 					 const void *buf, size_t len, int flags,
 					 unsigned zc_flags)
 {
-	io_uring_prep_rw(IORING_OP_SEND_ZC, sqe, sockfd, buf, (__u32) len, 0);
+	io_uring_prep_rw(IORING_OP_SEND_ZC, sqe, sockfd, buf, __io_uring_cap_len(len), 0);
 	sqe->msg_flags = (__u32) flags;
 	sqe->ioprio = zc_flags;
 }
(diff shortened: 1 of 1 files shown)


bitonic:
@axboe regarding branch or not branch: imo a predictable branch is better than a cmov (as in, it'll be cheaper in practice). But if you really care about having a cmov, I can make sure it compiles to that.

axboe:
I really think this should handle _all_ the cases where this is true. And this is very much using a branch, even if it's marked unlikely. A cap function ala:

```
size_t mask = -(len > UINT32_MAX);
return (unsigned int )((len & ~mask) | (UINT32_MAX & mask));
```

would be more efficient, and that help can just be used anywhere.

Also note the format of git commits for liburing. Needs a proper commit subject, commit message, identity, and signed-off-by line.

axboe:
Gotta run for now, I'll check back in tomorrow.

bitonic:
@axboe I've pushed the bit twiddly version. I've also amended all similar call sites I could find with a `__u32.*len` grep.

Should I just put the commit to be signed off by you?

bitonic:
@axboe is anything holding up the merge (apart from the commit message, which I'll amend if you confirm that I should put it as signed off by you)?

axboe:
Sorry, OOO for a bit, back next week. Only thing holding this back is I'm still pondering if we shouldn't just document the fact that the API len is a 32-bit unsigned. I'm not really convinced that it's worth adding code to cap the transfer length.

bitonic:
> Sorry, OOO for a bit, back next week. Only thing holding this back is I'm still pondering if we shouldn't just document the fact that the API len is a 32-bit unsigned. I'm not really convinced that it's worth adding code to cap the transfer length.

I think silently dropping the high 32 bits is terrible, documentation notwithstanding. `send` doesn't do that, so people will reasonably assume that the io_uring version will do the same, and it's a condition that will happen very rarely.

Are you concerned about performance? Capping the length will be immaterial compared to the cost of the place/pickup of the SQE onto the ring anyway.

src/include/liburing.h
```diff
@@ -919,17 +919,22 @@ IOURINGINLINE void io_uring_prep_statx(struct io_uring_sqe *sqe, int dfd,
 	sqe->statx_flags = (__u32) flags;
 }
 
+IOURINGINLINE __u32 __io_uring_cap_len(size_t len) {
```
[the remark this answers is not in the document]
bitonic: I've fixed both issues.

bitonic:
I think `io_uring_prep_fadvise`/`io_uring_prep_madvise`/`io_uring_prep_provide_buffers` are a lot more dubious (it should really have been a 64-bit length all along). But for `send`/`recv` there are basically no downsides.

src/include/liburing.h
```diff
@@ -919,17 +919,23 @@ IOURINGINLINE void io_uring_prep_statx(struct io_uring_sqe *sqe, int dfd,
...
+{
+	size_t mask = -(size_t)(len > UINT32_MAX);
+	return (len & ~mask) | (UINT32_MAX & mask);
+}
+
 IOURINGINLINE void io_uring_prep_fadvise(struct io_uring_sqe *sqe, int fd,
 					 __u64 offset, __u32 len, int advice)
 {
-	io_uring_prep_rw(IORING_OP_FADVISE, sqe, fd, NULL, (__u32) len, offset);
+	io_uring_prep_rw(IORING_OP_FADVISE, sqe, fd, NULL, __io_uring_cap_len(len), offset);
```
tavianator, line 931: `len` is a `__u32` here, so `__io_uring_cap_len` is a no-op

src/include/liburing.h
```diff
@@ -919,17 +919,23 @@ IOURINGINLINE void io_uring_prep_statx(struct io_uring_sqe *sqe, int dfd,
 	sqe->statx_flags = (__u32) flags;
 }
 
+static inline __u32 __io_uring_cap_len(size_t len)
+{
+	size_t mask = -(size_t)(len > UINT32_MAX);
+	return (len & ~mask) | (UINT32_MAX & mask);
```
tavianator, line 925: Based on https://godbolt.org/z/zGeqGGMeh, I'd suggest

```suggestion
	return len > UINT32_MAX ? UINT32_MAX : len;
```

src/include/liburing.h
```diff
@@ -919,17 +919,23 @@ IOURINGINLINE void io_uring_prep_statx(struct io_uring_sqe *sqe, int dfd,
...
-	io_uring_prep_rw(IORING_OP_FADVISE, sqe, fd, NULL, (__u32) len, offset);
+	io_uring_prep_rw(IORING_OP_FADVISE, sqe, fd, NULL, __io_uring_cap_len(len), offset);
 	sqe->fadvise_advice = (__u32) advice;
 }
 
 IOURINGINLINE void io_uring_prep_madvise(struct io_uring_sqe *sqe, void *addr,
 					 __u32 length, int advice)
 {
-	io_uring_prep_rw(IORING_OP_MADVISE, sqe, -1, addr, (__u32) length, 0);
+	io_uring_prep_rw(IORING_OP_MADVISE, sqe, -1, addr, __io_uring_cap_len(length), 0);
```
tavianator, line 938: Same here

src/include/liburing.h
bitonic: I had this back and forth with @axboe already. I think the cmov or branching version is better than the current version too (and moreover it'll leverage max instructions for architectures that have them, like aarch64). But since @axboe has stated that he prefers the bit twiddling version I'll let him decide.

bitonic: That's not how I understood https://github.com/axboe/liburing/pull/1436#issuecomment-3067197709 , given that it's replying to a comment that explicitly mentions cmov. But I'll let him comment.

bitonic:
@axboe any news? liburing is still silently broken (both in the docs and in the types) if you do large writes.

PR #1437, opened 2025-07-14
[GIT PULL] smp_load_acquire(ring->cq.khead) in __io_uring_peek_cqe() for consistent dereferencing

sohambagchi:
__io_uring_peek_cqe() is on the critical path for the user to consume completion-queue entries. It load_acquires the tail, loads the head, uses the head to index into the cq ring.

The C memory consistency model does not enforce ordering for address dependencies, like the one seen in:

```
LD1: unsigned head = *ring->cq.khead;
...
LD2: cqe = &ring->cq.cqes[(head & mask) << shift];
```
where LD2 is address-dependent on LD1. This means that LD2 could load a stale cqe due to LD1 being ordered after LD2. If instead, LD1 was a smp_load_acquire, the ordering would be enforced and correct cqes would be loaded.

Signed-off by: Soham Bagchi <soham.bagchi@utah.edu>
Co-authored by: Marco Elver <elver@google.com>
Co-authored by: Vijay Nagarajan <vijay@cs.utah.edu>
Co-authored by: Ryan Stutsman <stutsman@cs.utah.edu>

----
## git request-pull output:
```
The following changes since commit 2fd790493939c3b37688c2ddea4da9dcc436bb36:

  Merge tag 'gh-bot-musl-build-2025-07-14' of https://github.com/ammarfaizi2/liburing (2025-07-14 07:59:16 -0600)

  https://github.com/sohambagchi/liburing master

for you to fetch changes up to 81cb9479c8c409e3cf967c9be4094fdbe05187a5:

  smp_load_acquire(ring->cq.khead) in __io_uring_peek_cqe() for  consistent dereferencing (2025-07-14 16:22:02 -0400)

Soham Bagchi (1):
      smp_load_acquire(ring->cq.khead) in __io_uring_peek_cqe() for  consistent dereferencing

 src/include/liburing.h | 7 ++++++-
 1 file changed, 6 insertions(+), 1 deletion(-)
```
----

### DIFF
--- a/src/include/liburing.h
+++ b/src/include/liburing.h
@@ -1576,7 +1576,12 @@ IOURINGINLINE int __io_uring_peek_cqe(struct io_uring *ring,
 
 	do {
 		unsigned tail = io_uring_smp_load_acquire(ring->cq.ktail);
-		unsigned head = *ring->cq.khead;
+		
+		/**
+		 * A load_acquire on the head prevents reordering with the
+		 * cqe load below, ensuring that we see the correct cq entry.
+		 */
+		unsigned head = io_uring_smp_load_acquire(ring->cq.khead);
 
 		cqe = NULL;
 		available = tail - head;


axboe:
```IO_URING_READ_ONCE()``` should be enough here, no?

sohambagchi:
`IO_URING_READ_ONCE` is just a relaxed load operation, which has minimal guarantees in terms of ordering. Even address dependencies are not enforced by the C memory model, leading us to believe that the `IO_URING_READ_ONCE` would not be sufficient. 

The ordering guarantees are only applicable for releases, acquires and fences, hence we think the acquire is required on the head.

axboe:
Let me follow up the, is this an observed issue, or is it a purely theoretical one? Reason I ask is that any CQE posting should be happening from the task itself, and any switch of that task between CPUs always implies a full memory barrier. This should be true even for SQPOLL.

I guess it could potentially be possible for a signaled interruption, but that'd enter the kernel too to flush them out. Which again should imply full ordering on the memory side.

Maybe I'm missing something, in which case please enlighten me!

axboe:
I guess it could be possible for the case where you split submit and wait between separate threads. Which coincidentally is not something that's supported with eg DEFER_TASKRUN. For those cases I can see how this could be an observed issue. Side note on that - probably don't do that in general, it's not a good architecture for using io_uring. Doesn't mean it should result in a bug of course, it's just that I hate adding slowdowns for the oddball cases. Granted a relaxed read isn't a huge deal, so maybe I just need to swallow that one and move on :-)

PR #1446, opened 2025-08-01
DoS vulnerability in timeout handling on older kernels

sanjay20m:
Denial of Service (DoS) vulnerability in `liburing`'s timeout handling on older Linux kernels (those without `IORING_FEAT_EXT_ARG`).

The vulnerability occurs when an application uses a timeout function and a user-submitted operation happens to have the same `user_data` as the internal secret ID used for timeouts. In this scenario, the `_io_uring_get_cqe` function would return a success code while providing a `NULL` completion event pointer. This leads to a immediate NULL pointer dereference and crashes the application, resulting in a Denial of Service.

The patch modifies the `_io_uring_get_cqe` function in `src/queue.c`. It adds a crucial check to ensure that if a success code (`err >= 0`) is about to be returned, a valid completion event pointer (`cqe`) must exist. If not, the function now correctly returns `-EAGAIN` instead of a misleading success code, preventing the application crash.

### DIFF
--- a/src/queue.c
+++ b/src/queue.c
@@ -127,7 +127,9 @@ static int _io_uring_get_cqe(struct io_uring *ring,
 	} while (1);
 
 	*cqe_ptr = cqe;
-	return err;
+        if (err >= 0 && !cqe)
+        return -EAGAIN;
+        return err;
 }
 
 int __io_uring_get_cqe(struct io_uring *ring, struct io_uring_cqe **cqe_ptr,


axboe:
This is not a vulnerability, it's the application itself setting the user_data value... Not saying we should not make this bulletproof, but it's also just shooting yourself in the foot. It's not like you can cause others applications to crash.

axboe:
A few other comments:

1) The description of why this fix is necessary should go into the commit message. And the commit message should have a proper title too, "Update queue.c" means absolutely nothing.
2) Commit needs a signed-off-by
3) Do you have a test case? Given the dependency on EXT_ARG, it'd only really be something that'd hit on 5.10 or 5.4, everything else doesn't rely on internal timeouts.

axboe:
Closing due to lack of response.

PR #1453, opened 2025-09-03
test/recvsend_bundle: enlarge recv buf-ring to 2 MiB to avoid ENOBUFS…

schlad:
This is a minimal, test-only tweak to stabilize results across different arches without changing test logic.

IMHO, ideally we would find a way to allocate the buffer based on the system under test runtime socket SO_RCVBUF, so we would have a test without the hardcoded  RECV_BIDS and thus the buffer size.

So while I do not have a ready patch/solution, I wanted to see what do you think about it.

### DIFF
--- a/test/recvsend_bundle.c
+++ b/test/recvsend_bundle.c
@@ -21,7 +21,7 @@ static int nr_msgs;
 static int use_tcp;
 static int classic_buffers;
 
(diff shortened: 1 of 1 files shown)


axboe:
I think that work-around is likely just fine.

PR #1481, opened 2025-10-13
test: epwait: skip test_race on low CPU systems

schlad:
The test_race hangs on systems with 2 or fewer CPUs when using IORING_SETUP_SQPOLL, due to thread scheduling starvation. Specifically, the main thread and SQPOLL thread occupy both CPUs, leaving no CPU time for the writer thread to run, so on such systems the test_race should be skipped.

Today I was checking the test timeout (epwait) on openSuse Tumbleweed; after some checks I see the specific test which was failing was test_race; after pinning the writer thread and the sqpoll thread to specific cpus I could see the test started to work, so I assume the easiest solution is to make sure there is more than 2 cpus for this specific subtest.

### DIFF
--- a/test/epwait.c
+++ b/test/epwait.c
@@ -233,6 +233,7 @@ static void prune(struct epoll_event *evs, int nr)
 
 static int test_race(int flags)
 {
+	int nr_cpus;
 	struct io_uring_cqe *cqe;
 	struct io_uring_sqe *sqe;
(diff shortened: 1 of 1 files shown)


schlad:
I'll close this one. I think I'm off here. I managed to see a timeout on 4CPUs; will need to check this further. 
Sorry for the noise.

PR #1524, opened 2026-01-13
liburing: Add zone block device support

mannanal:
Add support for zone block devices (SMR/ZAC HDDs) to liburing. This enables io_uring based I/O operations on zoned storage devices.

The implementation provides:
- Zone discovery and information queries
- Zone reset, read, write operations
- Sequential zone write pointer management
- Batch and vectored I/O operations
- Zone filtering and search utilities

_Note: The zone support is conditionally compiled only when CONFIG_NOLIBC is disabled, as it requires standard C library functions (malloc, ioctl) that are not available in nolibc mode._

A simple test application (examples/zoned-simple-test.c) is included to demonstrate the API usage.

Example output:
  ```
=== liburing Zone Device Test ===
Device: /dev/sdcr
Write size: 256 KB

--- Discovering Zones ---
Found 104312 zones

--- Target Zone ---
Zone 1044:
  Type: SEQ_WRITE_REQ
  Condition: IMP_OPEN
  Start LBA: 547356672 (261.00 GB)
  Length: 524288 sectors (0.25 GB)
  Write Pointer: 547357184
  Available: 523776 sectors (0.25 GB)

--- Test Operations ---
1. Resetting zone...
   Zone reset successful (WP at LBA 547356672)
2. Writing 256 KB to zone...
   Write successful
3. Reading 256 KB from zone...
   Read successful
4. Verifying data integrity...
   PASSED: Data verified successfully

=== Test PASSED ===
```

Signed-off-by: Vishal Jose Mannanal <vishaljm@dropbox.com>

----

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -39,7 +39,8 @@ example_srcs := \
 	rsrc-update-bench.c \
 	proxy.c \
 	zcrx.c \
-	kdigest.c
+	kdigest.c \
+	zoned-test.c
 
 all_targets :=
 
--- a/examples/zoned-test.c
+++ b/examples/zoned-test.c
@@ -1,0 +1,281 @@
+/* SPDX-License-Identifier: MIT */
+/*
+ * Simple test application for liburing zone device support
+ * Tests basic zone operations on SMR/ZAC devices
+ */
+#define _GNU_SOURCE
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <errno.h>
+#include <sys/stat.h>
+#include <linux/blkzoned.h>
+#include <liburing.h>
+#include <liburing/zoned.h>
+
+#define TEST_BUFFER_SIZE (256 * 1024)  // 256KB
+#define IO_DEPTH 32
+
+static void print_usage(const char *prog)
+{
+    fprintf(stderr, "Usage: %s <device> [options]\n", prog);
+    fprintf(stderr, "Options:\n");
+    fprintf(stderr, "  -s <size>    Write size in KB (default: 256)\n");
+    fprintf(stderr, "  -z <zone>    Target zone index (default: first sequential zone)\n");
+    fprintf(stderr, "  -v           Verbose output\n");
+    fprintf(stderr, "\nExample: %s /dev/sdg\n", prog);
+}
+
+static int wait_completion(struct io_uring *ring, int expected)
+{
+    struct io_uring_cqe *cqe;
+    int completed = 0;
+    
+    while (completed < expected) {
+        int ret = io_uring_wait_cqe(ring, &cqe);
+        if (ret < 0) {
+            fprintf(stderr, "io_uring_wait_cqe: %s\n", strerror(-ret));
+            return ret;
+        }
+        
+        if (cqe->res < 0) {
+            fprintf(stderr, "I/O error: %s\n", strerror(-cqe->res));
+            io_uring_cqe_seen(ring, cqe);
+            return cqe->res;
+        }
+        
+        completed++;
+        io_uring_cqe_seen(ring, cqe);
+    }
+    
+    return 0;
+}
+
+int main(int argc, char *argv[])
+{
+    const char *device_path = NULL;
+    int fd, ret, opt;
+    struct io_uring ring;
+    struct uring_zone_info *zones = NULL;
+    size_t zone_count = 0;
+    size_t target_zone_idx = 0;
+    size_t write_size = TEST_BUFFER_SIZE;
+    bool verbose = false;
+    bool zone_specified = false;
+    void *write_buf = NULL, *read_buf = NULL;
+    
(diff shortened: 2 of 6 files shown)


mannanal:
CC: @damien-lemoal @yhr io_uring changes enabling ZBD support.

damien-lemoal:
See the comment I sent on your fio PR. It applies here as well.

mannanal:
> See the comment I sent on your fio PR. It applies here as well.

Just copying here the same comment for reference.

Thanks for looking into the PR. You're right that io_uring doesn't guarantee ordering, which can be an issue for sequential zones. I'm working with SMR HDDs and making sure only one write is sent to a zone at a time. This is similar to how fio handles zonemode=zbd with io_uring, using user-space logic to avoid write pointer violations. So far this has worked well, but I agree that kernel-side ordering would make things more robust.

That said, this is not unique to io_uring, where the same applies to libaio or any other async I/O mode. As long as only one write is in flight per zone, and the write pointer is respected, it works reliably. io_uring just brings additional performance benefits like lower syscall overhead, fewer context switches, and efficient scaling across zones.

We’ve run several days of tests with SMR drives using this model and haven’t seen any issues so far. I'll be happy to compare notes if you've observed otherwise.

damien-lemoal:
That is not correct. libaio is fine for writing a zone at QD>1. Accidentally so, but there are no issues.
For your patches, it would be nice to enforce the QD=1 restriction for writes, and also check that IO splitting does not create issues even at QD=1 with super large write commands.

mannanal:
> That is not correct. libaio is fine for writing a zone at QD>1. Accidentally so, but there are no issues.
> For your patches, it would be nice to enforce the QD=1 restriction for writes, and also check that IO splitting does not create issues even at QD=1 with super large write commands.

Thanks Damien, good to know about libaio’s behavior. That helps clarify why it’s held up in practice with zoned devices. 

For io_uring, I’ll enforce QD=1 per zone explicitly and add a check to avoid large writes that could trigger internal bio splitting. Appreciate the heads-up on that corner case.

espoal:
What does this PR enable in particular? What cannot be done today?

I'm concerned because I've been using zoned devices with io_uring for years, was I doing something wrong?

axboe:
Honestly I think this should be in a library that just depends on liburing, not in liburing itself. It's fairly specific building blocks that just sit on top of liburing, it's not really functionality that liburing needs to provide.

damien-lemoal:
> Honestly I think this should be in a library that just depends on liburing, not in liburing itself. It's fairly specific building blocks that just sit on top of liburing, it's not really functionality that liburing needs to provide.

I agree. As is, io_uring works just fine to read and write sequential zones, modulo the QD=1 write per zone limitation . So does fio as well with zonemode=zbd. That is usable with ioengine=io_uring. No patches are needed for doing media-access IOs. For zone reporting and zone management, we could add io_uring operations, but since that is not really the fast path generally, I am not convinced it is worth the effort.

If this change facilitates coding some application, making it a separate library is fine with me.

axboe:
Let's close this one up for now.

mannanal:
> What does this PR enable in particular? What cannot be done today?
> 
> I'm concerned because I've been using zoned devices with io_uring for years, was I doing something wrong?

Thanks @espoal for the review. This PR adds zone block device helper functions to liburing, making it easier to work with SMR/ZNS devices.  It provides zone discovery (wrapping BLKREPORTZONE ioctl), zone reset operations, and convenience wrappers that check write pointer positions before submitting I/O.

The helpers don't enforce ordering or QD=1 constraints and applications are still responsible for ensuring only one write is in-flight per sequential zone and managing submission order. This simplifies zone operations but doesn't enforce constraints.

If you've been using zoned devices with io_uring successfully, this just makes that easier without adding new safety mechanisms.

_I stand corrected on my earlier comment as the library doesn't enforce QD=1, it only provides helper functions for zone operations. Applications are still responsible for managing write ordering and queue depth. Hence editing my response, thanks._

mannanal:
> Honestly I think this should be in a library that just depends on liburing, not in liburing itself. It's fairly specific building blocks that just sit on top of liburing, it's not really functionality that liburing needs to provide.

That’s a good point.  I agree this logic could live in a separate wrapper. My thinking was that zoned devices, particularly SMR HDDs, are sensitive to subtle misuses like QD > 1 to a sequential zone or large writes triggering splits. Putting basic zone-safe patterns in liburing would help guide users away from silent data correctness issues without requiring every application to rebuild that logic. Leaving it to external libraries would work too (say liburing-zbd), but it adds overhead for apps to get things right, especially since io_uring doesn’t provide ordering guarantees today. 

That said, I’m happy to move this into a thin wrapper library if you think liburing itself should stay focused on pure I/O mechanics.

axboe:
I do think it's useful code, and it should go somewhere so that people can use it. I just don't think liburing is that spot. However, I am open to stuffing it into examples/ and having code in there that shows you exactly what to do. I just don't think it needs to be wired up in the library itself.

PR #1539, opened 2026-02-18
tests: Add a helper to check the minimum required kernel version

schlad:
io-wq-exit was added to liburing before the corresponding fix landed in the mainline kernel tree. Add a helper that checks whether the running kernel meets a minimum version and returns T_EXIT_SKIP if not.

Proposal to add a small helper which should make it easier to run tests in - for instance - any CI.
If you agree, we could slowly go towards some more formal structure of the tests, so that - for instance - sub-tests could emit some clear t1, t2 ... tx etc. 
That way it might be easier to run the tests and quickly process any failures etc.

### DIFF
--- a/test/helpers.c
+++ b/test/helpers.c
@@ -10,6 +10,9 @@
 #include <unistd.h>
 #include <stdarg.h>
 #include <sys/types.h>
+#include <sys/utsname.h>
+#include <limits.h>
+#include <errno.h>
 
 #include <arpa/inet.h>
 #include <netinet/ip.h>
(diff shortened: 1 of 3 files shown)


axboe:
I don't think we should do this, it's better to just have it fail, when it should fail. This also doesn't work on stable kernels, which may also have the bug but may need eg 6.12.139. I'd rather not keep track of these things. Yes it's sometimes a bit annoying if a test case gets added early, but at the end of the day, if the kernel is missing a fix it's missing a fix. Better to flag that than ignore it.

schlad:
> I don't think we should do this, it's better to just have it fail, when it should fail. This also doesn't work on stable kernels, which may also have the bug but may need eg 6.12.139. I'd rather not keep track of these things. Yes it's sometimes a bit annoying if a test case gets added early, but at the end of the day, if the kernel is missing a fix it's missing a fix. Better to flag that than ignore it.

fair point.

PR #1544, opened 2026-03-01
nolibc: fix integer overflow in __uring_malloc

rootvector2:
The addition of sizeof(*heap) and len can overflow size_t when len is
close to SIZE_MAX. This results in an undersized mmap allocation while
heap->len records the small wrapped value, leading to a heap buffer
overflow on subsequent writes to the returned pointer.

Add an overflow check before passing the computed size to mmap.

Signed-off-by: ```Naveed <dxbnaveed.k@gmail.com>```

### DIFF
--- a/src/nolibc.c
+++ b/src/nolibc.c
@@ -33,13 +33,18 @@ struct uring_heap {
 void *__uring_malloc(size_t len)
 {
 	struct uring_heap *heap;
+	size_t total = sizeof(*heap) + len;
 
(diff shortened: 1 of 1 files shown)


axboe:
You're missing a ```Signed-off-by``` in your commit message.

axboe:
Pulling this as a cleanup, there's no real issue here in terms of the liburing use cases. Thanks!

PR #1577, opened 2026-05-17
harden ring sizing arithmetic against overflow

metsw24-max:
setup: harden ring sizing arithmetic against overflow

Several setup-time size calculations relied on unchecked arithmetic
 for `count * sizeof(...)` products and shift-expanded entry counts.
 Large or malformed inputs could wrap these calculations and produce
 undersized mappings or allocations.

Introduce `__size_mul()` as a centralized checked multiplication
 helper and route all `mmap()/munmap()-relevant` sizing through it.
Convert the internal sizing helpers to explicit error-returning
 interfaces so overflow can be propagated consistently.

Also fix the shift-before-multiply overflow class by promoting
 counts to `size_t` before applying `SQE/CQE` size shifts.

Add regression coverage for both direct helper boundary behavior
 and end-to-end oversized `io_uring_setup_buf_ring()` handling,
including verification that failed setup attempts do not leave
 behind partial registration state.

Signed-off-by: Metsw Max [metsw24-max@gmail.com](mailto:metsw24-max@gmail.com)

### DIFF
--- a/src/setup.c
+++ b/src/setup.c
@@ -95,26 +95,36 @@ void io_uring_setup_ring_pointers(struct io_uring_params *p,
 	cq->ring_entries = *cq->kring_entries;
 }
 
-static size_t params_sqes_size(const struct io_uring_params *p, unsigned sqes)
+static int params_sqes_size(const struct io_uring_params *p, unsigned sqes,
+			    size_t *out)
 {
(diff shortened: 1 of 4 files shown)


axboe:
Explanation goes into the commit message (which is basically empty), the github parts don't matter one bit. But this one has been covered before - there's no risk of overflow with the sizes supported, so let's not overcomplicate anything.

PR #1578, opened 2026-05-18
register preserve existing ring mappings on resize remap failure

uwezkhan:
io_uring_resize_rings currently unmaps the existing SQ and CQ mappings before attempting to create the replacement mappings.

If the remap step fails the ring is left with cleared internal state and later operations such as io_uring_submit io_uring_get_sqe and io_uring_wait_cqe can dereference NULL ring pointers and crash.

This change makes the resize remap flow transactional by creating the new mappings first and only replacing the existing mappings after successful completion.

The success path behavior remains unchanged while resize failures now preserve a valid operational ring state.

The patch is intentionally minimal and localized to the resize path with no API or ABI changes.

### DIFF
--- a/src/register.c
+++ b/src/register.c
@@ -451,6 +451,8 @@ int io_uring_register_ifq(struct io_uring *ring,
 
 int io_uring_resize_rings(struct io_uring *ring, struct io_uring_params *p)
 {
+	struct io_uring_sq sq;
+	struct io_uring_cq cq;
 	unsigned sq_head, sq_tail;
 	int ret;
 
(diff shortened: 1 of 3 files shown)


axboe:
Please put the commit message in the, you know, _actual_ commit. It serves no purposes on just github, it needs to go into git history. You also need a ```Signed-off-by``` line. When in doubt, check existing commits for how things should be properly formatted.

How did you test this?

uwezkhan:
updated the actual git commit message and added the signed-off-by line.

i also added a regression test for the failed remap path during io_uring_resize_rings(). The test forces mmap failure using RLIMIT_AS constraints and verifies that the original ring remains usable after the resize failure instead of being left in an invalid state.

axboe:
Please squash the fix into one commit, you have duplicate commits now. The test case can be a separate commit - thanks for adding that, btw!

uwezkhan:
oops i accidentally ended up with duplicate fix commits while reworking the commit message and wording.

I have squash the resize fix into a single commit and kept the regression test as a separate commit.

axboe:
Please check again, you have even more commits in there now.

uwezkhan:
**Cleaned it up now, only the fix commit and the test commit are left in the branch.**

axboe:
Thanks!

PR #1581, opened 2026-05-26
queue: fix NULL deref in _io_uring_get_cqe() for EXT_ARG_REG path

### LINKED ISSUE #1567: Memory access violation when using io_uring_submit_and_wait_reg()
I'm sorry for the mistakes I'm probably gonna make in this issue in advance, as I'm currently learning C and I'm still very new to Linux programming, in general.

When trying to use the `io_uring_submit_and_wait_reg()` function, given a properly setup timer and zero submitted SQEs, everything works fine. But as soon as I try to submit SQEs, I get a SIGSEGV error, specifically when not enough (or zero) SQEs have been completed before the deadline, triggering the timeout. It took me a good while to figure out what's wrong, but after stepping through in the debugger, I've found that a particular branch in the internal `_io_uring_get_cqe()` function tries to access memory that is... not memory at all. I'm not sure as to the specifics of this issue, but I was able to find that the culprit was this line:
```c
// src/queue.c:106
		if (looped && data->has_ts) {
			struct io_uring_getevents_arg *arg = data->arg;

			if (!cqe && arg->ts && !err) // <== arg->ts is illegal here
				err = -ETIME;
			break;
		}
```
After some further digging up and down the call stack, I've found that the `data->arg` member field in this particular case is set __not__ to an address, but to an offset (i.e. regular non-address integer):
```c
// src/queue.c:338
int io_uring_submit_and_wait_reg(struct io_uring *ring,
				 struct io_uring_cqe **cqe_ptr,
				 unsigned wait_nr, int reg_index)
{
	unsigned long offset = reg_index * sizeof(struct io_uring_reg_wait);

	struct get_data data = {
		.submit		= __io_uring_flush_sq(ring),
		.wait_nr	= wait_nr,
		.get_flags	= IORING_ENTER_EXT_ARG | IORING_ENTER_EXT_ARG_REG,
		.sz		= sizeof(struct io_uring_reg_wait),
		.has_ts		= true,
		.arg		= (void *) (uintptr_t) offset, // <== Here's the offset being casted to an address
	};

	if (!(ring->features & IORING_FEAT_EXT_ARG))
		return -EINVAL;

	return _io_uring_get_cqe(ring, cqe_ptr, &data);
}
```
I was able to "fix" this by just removing the `arg->ts` check, but this most assuredly wrong:
```c
// src/queue.c:106
		if (looped && data->has_ts) {
			struct io_uring_getevents_arg *arg = data->arg;

			if (!cqe && !err) // <== arg->ts is removed
				err = -ETIME;
			break;
		}

```
Is it indeed a bug, or is my liburing-jutsu lacking?

XananasX7:
When io_uring_submit_and_wait_reg() is used, data->arg holds a register
offset cast to void *, not a pointer to io_uring_getevents_arg. If no CQE
is available on the first pass, the looped && has_ts branch dereferences
data->arg as a struct pointer, which causes a segfault.

Fix this by checking IORING_ENTER_EXT_ARG_REG in get_flags before the
dereference. For the registered-wait path the kernel handles the timeout,
so return -ETIME directly without touching data->arg.

Fixes #1567.

### DIFF
--- a/src/queue.c
+++ b/src/queue.c
@@ -101,10 +101,23 @@ static int _io_uring_get_cqe(struct io_uring *ring,
 		if (!need_enter)
 			break;
 		if (looped && data->has_ts) {
-			struct io_uring_getevents_arg *arg = data->arg;
+			/*
+			 * When IORING_ENTER_EXT_ARG_REG is set, data->arg
+			 * carries a register-wait offset (an integer), not a
+			 * pointer to io_uring_getevents_arg.  Dereferencing it
+			 * as a struct pointer causes a memory access violation.
+			 * For the registered-wait path the kernel enforces the
+			 * timeout, so treat any timeout the same as -ETIME here.
+			 */
+			if (data->get_flags & IORING_ENTER_EXT_ARG_REG) {
+				if (!cqe && !err)
+					err = -ETIME;
+			} else {
+				struct io_uring_getevents_arg *arg = data->arg;
 
-			if (!cqe && arg->ts && !err)
-				err = -ETIME;
+				if (!cqe && arg->ts && !err)
+					err = -ETIME;
+			}
 			break;
 		}
 


axboe:
Looks reasonable, but please follow the commit guide for liburing. Basically everything you put in here needs to go in the commit message, but please keep it less LLM and more normal reading (and shorter). And you need a ```Signed-off-by``` line with a real identity + email.

XananasX7:
Thanks for the feedback. I've rewritten the commit message to be shorter and plain — no more verbose structure, and the key info is in the body now. Also added the Signed-off-by line. PR description updated to match.

axboe:
Thanks, that looks much better. I'll run this through some testing.

PR #1583, opened 2026-05-28
recvmsg: compute cmsg nexthdr end without firsthdr

rootvector2:
Noticed io_uring_recvmsg_cmsg_nexthdr() computes the cmsg region end as
io_uring_recvmsg_cmsg_firsthdr(o, msgh) + o->controllen. firsthdr
returns NULL when o->controllen < sizeof(struct cmsghdr), so this
becomes a non-zero offset applied to a null pointer, which is undefined
behavior. Compute end directly as name + namelen + controllen so the
arithmetic is always on a real pointer; the value matches the old
expression when firsthdr would have succeeded.

### DIFF
--- a/src/include/liburing.h
+++ b/src/include/liburing.h
@@ -1237,8 +1237,8 @@ io_uring_recvmsg_cmsg_nexthdr(struct io_uring_recvmsg_out *o, struct msghdr *msg
 
 	if (cmsg->cmsg_len < sizeof(struct cmsghdr))
 		return NULL;
-	end = (unsigned char *) io_uring_recvmsg_cmsg_firsthdr(o, msgh) +
-		o->controllen;
+	end = (unsigned char *) io_uring_recvmsg_name(o) +
+		msgh->msg_namelen + o->controllen;
 	cmsg = (struct cmsghdr *)((unsigned char *) cmsg +
 			CMSG_ALIGN(cmsg->cmsg_len));
 


Kprateek283:
Just curious, did you actually hit a crash with this in the production, or was it flagged by ubsan/a static analyzer?

rootvector2:
not a production crash, it was flagged by static analysis. when `o->controllen < sizeof(struct cmsghdr)`, `firsthdr` returns NULL and `end` ends up as `NULL + o->controllen`, which is UB. computing it as `name + namelen + controllen` gives the same value whenever firsthdr would've returned non-NULL, just without the null-base arithmetic.

axboe:
When can ```o->controllen < sizeof(struct cmsghdr)``` occur, then?

rootvector2:
in the documented `firsthdr` then `nexthdr` loop it can't. `nexthdr` is only reached after `firsthdr` returned non-NULL, so `o->controllen >= sizeof(struct cmsghdr)` already holds and the `+ o->controllen` is on a real pointer. the analyzer trips on it because `nexthdr` calls `firsthdr` again on its own, and in isolation that call can return NULL before the offset is applied.

so it's really a cleanup: dropping the redundant `firsthdr` and computing `end` from `name + namelen + controllen` gives the identical value without the null-base arithmetic. no behavior change in the normal flow.

rootvector2:
any update?

PR #1587, opened 2026-06-01
liburing.h: fix integer overflow in recvmsg_cmsg_nexthdr

rootvector2:
Noticed recvmsg_cmsg_nexthdr advances cmsg by CMSG_ALIGN(cmsg_len) before any bounds check. A cmsg_len near SIZE_MAX makes the align wrap to a small value, or the pointer add wrap past the control buffer, so the '> end' checks pass and the wild cmsg gets returned and dereferenced. Bound the aligned length against the remaining control space before advancing, the same way the recent recvmsg_validate fix did.


axboe:
Closing due to build failures.

PR #1588, opened 2026-06-01
liburing.h: reject negative op in io_uring_opcode_supported

rootvector2:
io_uring_opcode_supported() guards only the upper bound, so a negative op slips past it and p->ops[op] reads before the probe allocation - op == -3 reads 8 bytes ahead of the 16-byte struct, which ASan reports as a heap-buffer-overflow. Add the op < 0 check. Full context is in the commit message.


axboe:
Same as the other one, this is an API misuse in the application, strictly a cleanup for apps that choose to shoot themselves in the foot.

PR #1589, opened 2026-06-01
examples/io_uring-udp: compute buffer size and offset in size_t

rootvector2:
buffer_size() and get_buffer() take their shift count from the -b option but evaluate the shift in 32-bit before widening, so a large value is undefined behavior and otherwise truncates the result. Do both shifts in size_t. Building with -fsanitize=undefined and running with -b 40 flags the old code (shift exponent 40 too large for 32-bit type), and buffer_size() returns 256 instead of 2^40.

### DIFF
--- a/examples/io_uring-udp.c
+++ b/examples/io_uring-udp.c
@@ -36,12 +36,12 @@ struct ctx {
 
 static size_t buffer_size(struct ctx *ctx)
 {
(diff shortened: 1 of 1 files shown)


axboe:
Pulling this as a cleanup, in practice there are no issues here.

PR #1591, opened 2026-06-04
ucontext-cp: don't free coroutine stack and context in use

rootvector2:
`copy_file_wrapper` runs as a `makecontext` coroutine on `pctx->stack_buf`, but on completion it frees that stack and `pctx` and then `swapcontext`s through them, so the final switch back to main runs on freed memory. found it reading the cleanup path. move the frees of `stack_buf` and `pctx` into `main`, which reclaims them once the coroutine has finished.

### DIFF
--- a/examples/ucontext-cp.c
+++ b/examples/ucontext-cp.c
@@ -164,10 +164,13 @@ static void copy_file_wrapper(arguments_bundle *pbundle)
 	free(iov.iov_base);
 	close(pbundle->infd);
 	close(pbundle->outfd);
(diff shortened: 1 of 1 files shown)


ammarfaizi2 (changes requested):
Please, read the pull request guidelines.

 - The commit message must contain a Signed-off-by tag.
 - The explanation about the commit should be included in the commit message itself, word-wrapped at 72 chars.

rootvector2:
done. added the sign-off and moved the explanation into the commit message body, wrapped at 72.

PR #1594, opened 2026-06-06
test: min-timeout: Call io_uring_queue_exit

chrehrhardt:
Call io_uring_queue_exit after each test to avoid request completions after the end of the test.


chrehrhardt:
Removed unrelated whitespace change.

axboe:
Previous version already merged.

Issue #1596, opened 2026-06-08
Proposal: Add minimal multishot-echo-server example


Kprateek283:
Hi,
I noticed there isn't a simple tutorial example for modern io_uring networking. Beginners trying to learn how to combine Multishot Accept, Multishot Recv, and Buffer Rings are currently forced to reverse-engineer examples/proxy.c , which is highly complex (~2,500 lines) and covers advanced features like splice and zero-copy.
Would you be open to a PR adding a minimal multishot-echo-server.c? It would serve as a clean "Hello World" demonstrating just a basic event loop using those three features, with proper buffer lifecycle management.
I'd be happy to implement and submit this if it fits the repository's goals.

axboe:
Yep, I'd definitely be open to that. Proxy is definitely a bit on the complicated side, because it tries to show all of the variants. At some point I might cut it down a bit. But I think there's certainly room for a basic echo server that shows all the best practices, so feel free to submit that for ```examples/```.

PR #1597, opened 2026-06-09
examples: add minimal multishot echo server

Kprateek283:
Add a minimal multishot echo server example demonstrating:

- Multishot accept
- Multishot receive
- Provided buffer rings

This provides a simpler networking example than proxy.c for developers learning modern io_uring APIs.
Related: #1596

### DIFF
--- a/examples/echo-server.c
+++ b/examples/echo-server.c
@@ -1,0 +1,447 @@
+/* SPDX-License-Identifier: MIT */
+/*
+ * Simple TCP echo server demonstrating modern io_uring networking best
+ * practices:
+ *
+ *   - Multishot accept (arm once, get a CQE per new connection)
+ *   - Multishot receive (arm once per connection, get a CQE per data chunk)
+ *   - Provided buffer rings (kernel picks buffers from a shared pool)
+ *   - Proper buffer lifecycle management (recycle after send completes)
+ *
+ * Requires kernel >= 6.0 for buffer rings and multishot recv.
+ * DEFER_TASKRUN (kernel >= 6.1) is used when available, with automatic
+ * fallback to COOP_TASKRUN on older kernels.
+ *
+ * Usage: ./echo-server [port]
+ *        Default port: 8000
+ *
+ * Test with: nc localhost 8000
+ */
+#include <stdio.h>
+#include <stdlib.h>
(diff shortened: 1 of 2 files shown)


examples/echo-server.c
```diff
@@ -0,0 +1,449 @@
...
+ * multishot accept arms once and generates a CQE for each new incoming
+ * connection. It stays active until an error occurs (signaled by the
+ * absence of IORING_CQE_F_MORE in cqe->flags).
+ */
+static void add_multishot_accept(void)
+{
+	struct io_uring_sqe *sqe;
+
+	sqe = get_sqe();
+	if (!sqe)
```
axboe: These should probably be fatal conditions, not silently ignored. There are a few of those. It's really a "this should never happen" scenario, hence could be an assert() too.

examples/echo-server.c
```diff
@@ -0,0 +1,449 @@
...
+			continue;
+		if (ret < 0) {
+			fprintf(stderr, "submit_and_wait: %s\n",
+					strerror(-ret));
+			break;
+		}
+
+		count = 0;
+		io_uring_for_each_cqe(&ring, head, cqe) {
+			switch (decode_type(cqe->user_data)) {
```
axboe, line 364: Should probably have a ```default:``` error case?

axboe:
Just a few minor comments, overall looks nice and simple and agree this would be a good example to add!

examples/echo-server.c
Kprateek283: Agreed. I've updated it to be a fatal error via `exit(1)`. I chose `exit(1)` over `assert()` to ensure the failure is still caught and the error message is printed even when the example is compiled with `-DNDEBUG`.

examples/echo-server.c
Kprateek283: You're right. Added a default: case with exit(1). An unknown event type should be impossible, so this is a fatal logic error rather than a recoverable condition.

axboe:
Can you squash the two commits? Mostly just so that intermediate state doesn't show up in the tree, but also because you didn't add your SOB to commit 2.

Kprateek283:
Done. Please let me know if it requires any further input from my side.

PR #1598, opened 2026-06-10
test/poll-update-trigger: fix poll mask byte order

schlad:
sqe->poll32_events is "word-reversed for BE", so assigning POLLOUT directly gives the kernel the wrong mask on s390 and the test hangs. Use io_uring_prep_poll_update() which handles the byte swap correctly.

Tested on s390;
before the test: 
`Running test poll-update-trigger.t Test poll-update-trigger.t timed out (may not be a failure)`
With the patch the test passes on s390.


axboe:
Thanks!

PR #1599, opened 2026-06-11
Harden CI workflow security

Alb3e3:
This tightens the CI workflow security posture without changing the build matrix itself.

- set the workflow token to read-only repository contents access
- avoid direct GitHub context interpolation inside the shell command that builds the commit list
- pass generated matrix commit values through environment variables and quote them before checkout

I verified the workflow YAML still parses, `git diff --check` passes, and `zizmor` no longer reports the previous excessive-permissions or template-injection findings for this workflow. The remaining `zizmor` findings are action pinning warnings, which I left out of this focused change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,9 @@ on:
   push:
   pull_request:
 
+permissions:
+  contents: read
+
 
 jobs:
   get_commit_list:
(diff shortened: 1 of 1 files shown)


ammarfaizi2 (changes requested):
I am okay with the changes, but as always, please follow the pull request guidelines.

1) The explanation should be included in the commit message and word-wrapped at 72 characters.
2) The commit must contain a `Signed-off-by` tag.

Alb3e3:
Thanks for the review. I've addressed both points in the latest push: the explanation is now in the commit message word-wrapped at 72 characters, and the commit carries a Signed-off-by tag. Please take another look when you have a moment.

PR #1601, opened 2026-06-15
man: document that send/recv can see -EAGAIN without MSG_DONTWAIT

tchaikov:
The send/recv prep man pages imply -EAGAIN is only expected when the application sets MSG_DONTWAIT. In practice io_uring can also complete a request that did not set MSG_DONTWAIT with -EAGAIN: when it cannot make progress, it stops retrying rather than waiting indefinitely (see kernel commit c16bda37594f "io_uring/poll: allow some retries for poll triggering spuriously", which bounds the retry attempts).

Applications submitting blocking-style socket sends/recvs (no MSG_DONTWAIT) can therefore observe -EAGAIN under load and must treat it as transient and reissue. Note this in the ERRORS section of the send, sendmsg, recv and recvmsg prep pages.

----
## git request-pull output:
```
The following changes since commit 2750b94a77ea9e4e156fd0caa5c14552e7d1397f:

  Merge branch 'poll-update-trigger' of https://github.com/schlad/liburing (2026-06-10 06:57:31 -0600)

  git@github.com:tchaikov/liburing.git man-document-eagain-without-dontwait

for you to fetch changes up to f4cdd7d3fe004d1315bd367fefbc189aa92511de:

  man: document that send/recv can see -EAGAIN without MSG_DONTWAIT (2026-06-15 20:06:25 +0800)

Kefu Chai (1):
      man: document that send/recv can see -EAGAIN without MSG_DONTWAIT

 man/io_uring_prep_recv.3    | 17 ++++++++++++++++-
 man/io_uring_prep_recvmsg.3 | 19 ++++++++++++++++++-
 man/io_uring_prep_send.3    | 17 ++++++++++++++++-
 man/io_uring_prep_sendmsg.3 | 16 ++++++++++++++++
 4 files changed, 66 insertions(+), 3 deletions(-)
```
----
<details>

## Pull Request Guidelines
```
```

2. Then an empty line.
```
```

certain specific format).

of truth, not github.

URL.

```
```

Use a real email address!

### Commit message example:
```

safe without synchronization.

```

</details>

----

### DIFF
--- a/man/io_uring_prep_recv.3
+++ b/man/io_uring_prep_recv.3
@@ -135,7 +135,28 @@ Instead it returns the negated
 .I errno
 directly in the CQE
 .I res
-field.
+field. Some common error cases are:
+.TP
+.B -EAGAIN
+The operation would have blocked. io_uring normally arms internal poll and
+reissues the request once the socket is ready, so this does not occur for a
+socket in blocking mode. It is returned when the operation is non-blocking and
+cannot complete immediately, that is when
+.B MSG_DONTWAIT
+was set in the
+.I flags
+argument or the socket is marked
+.BR O_NONBLOCK ;
+.BR recv (2)
+treats the two equivalently. To have io_uring poll and retry instead, do not
+mark the socket
+.B O_NONBLOCK
+and do not set
+.BR MSG_DONTWAIT .
+Otherwise treat
+.B -EAGAIN
+as transient and reissue the request.
+.P
 .SH NOTES
 Despite accepting a size_t number of bytes, these functions can transfer at most
 INT_MAX bytes per call (the maximum for the underlying syscall interface).
--- a/man/io_uring_prep_recvmsg.3
+++ b/man/io_uring_prep_recvmsg.3
@@ -107,7 +107,30 @@ Instead it returns the negated
 .I errno
 directly in the CQE
 .I res
-field.
+field. Some common error cases are:
+.TP
+.B -EAGAIN
+The operation would have blocked. io_uring normally arms internal poll and
+reissues the request once the socket is ready, so this does not occur for a
+socket in blocking mode. It is returned when the operation is non-blocking and
+cannot complete immediately, that is when
+.B MSG_DONTWAIT
+was set in the
+.I msg_flags
+of the
+.I msghdr
+argument or the socket is marked
+.BR O_NONBLOCK ;
+.BR recv (2)
+treats the two equivalently. To have io_uring poll and retry instead, do not
+mark the socket
+.B O_NONBLOCK
+and do not set
+.BR MSG_DONTWAIT .
+Otherwise treat
+.B -EAGAIN
+as transient and reissue the request.
+.P
 .SH NOTES
 As with any request that passes in data in a struct, that data must remain
 valid until the request has been successfully submitted. It need not remain
--- a/man/io_uring_prep_send.3
+++ b/man/io_uring_prep_send.3
@@ -184,7 +184,28 @@ Instead it returns the negated
 .I errno
 directly in the CQE
 .I res
-field.
+field. Some common error cases are:
+.TP
+.B -EAGAIN
+The operation would have blocked. io_uring normally arms internal poll and
+reissues the request once the socket is ready, so this does not occur for a
+socket in blocking mode. It is returned when the operation is non-blocking and
+cannot complete immediately, that is when
+.B MSG_DONTWAIT
+was set in the
+.I flags
+argument or the socket is marked
+.BR O_NONBLOCK ;
+.BR send (2)
+treats the two equivalently. To have io_uring poll and retry instead, do not
+mark the socket
+.B O_NONBLOCK
+and do not set
+.BR MSG_DONTWAIT .
+Otherwise treat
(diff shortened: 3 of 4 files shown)


axboe:
This is essentially a "will never happen in an app" case, but added so that malicious/odd reproducers don't get stuck forever. Curious if this is something you actually ran into, or whether you just skimmed the source and spotted this?

tchaikov:
@axboe hi Jens,

it's real, not from skimming the source. actually, it showed up when testing Ceph's crimson test suite, which runs on Seastar's io_uring reactor backend, on a riscv64 box. the socket test `test_preemptive_down` started aborting intermittently under load. the app only expects `broken_pipe`/`connection_reset` when a connection goes away, so `EAGAIN` tripped a `abort()`.

it only shows up under heavy CPU contention. the test runs Seastar on 2 cores (`--smp 2`), and we pin everything to those same 2 cores while running busy background process there too, so the reactor is heavily oversubscribed:
```bash
for i in $(seq 12); do 
  taskset -c 0-1 yes >/dev/null & 
done
taskset -c 0-1 ./unittest-seastar-socket --reactor-backend io_uring --smp 2
```

With the cores that contended, 2 of 3 runs abort. ftrace on the io_uring events caught the path:
```
submit           SENDMSG
poll_arm         SENDMSG                 # buffer full, poll armed
complete         SENDMSG result 34199    # partial send
submit           SENDMSG                 # retry remainder, REQ_F_POLLED
queue_async_work SENDMSG
iou-wrk complete SENDMSG result -11
```

a partial send that can't make further progress while the reactor is starved, hitting the retry cap from kernel commit c16bda37594f. So yes, pathological scheduling, but a real workload rather than a crafted reproducer.

we fixed it in Seastar by treating `EAGAIN` as transient and retrying through poll (20 passes out 20 runs after that). the man page note just gives the next person who hits a stray EAGAIN under load somewhere to land, instead of assuming io_uring is broken. if you'd rather not document a case that won't occur under normal scheduling, i am happy to drop or reword it.

avikivity:
Why is it different from ordinary blocking sendmsg?

Having very rare EAGAIN means most users will get it wrong, since it won't be exercised in tests. It would be better to be more similar to the sendmsg syscall.

axboe:
This looks like a bug, because why is iou-wrk completing it with ```-EAGAIN```? It's a blocking issue at that point because we had to completely give up. Is ```MSG_DONTWAIT``` set there?

THAT looks like the thing that needs fixing, not the documentation. Let me know if you're up for debugging that, I'm unfortunately about to head OOO for a week or so. If you have a reproducer I can run locally, then let me know, may be able to do something about that today.

axboe:
> Why is it different from ordinary blocking sendmsg?

It should not be, looks like that's the bug.

axboe:
Followup question - are you using ```SO_SNDTIMEO```?

axboe:
And what kernel are you testing on, that's crucial information that should be included in here too.

axboe:
Here's an example of ```sock_sendmsg``` returning ```-EAGAIN``` with ```SO_SNDTIMEO``` even when ```MSG_DONTWAIT``` isn't set:

```
[   54.070197] send-eagain.t: ret=-11, msg_flags=4040, nowait=1
[   54.070201] send-eagain.t: ret=-11, msg_flags=4040, nowait=1
[   54.070205] send-eagain.t: ret=-11, msg_flags=4040, nowait=1
[   54.070210] send-eagain.t: ret=-11, msg_flags=4040, nowait=1
[   54.070214] send-eagain.t: ret=-11, msg_flags=4040, nowait=1
[   54.176177] iou-wrk-948: ret=-11, msg_flags=4000, nowait=0
```

You can see it being punted to io-wq at the end, which then correctly does a ```!MSG_DONTWAIT``` issue, yet you still get ```-EAGAIN``` back. The exact same thing should happen with a normal ```send(2)``` or ```sendmsg(2)```.

Yes, this is influenced by the retry cap, but that's also what ensures that this loop eventually stops in the kernel.

Now I'm curious if you're using ```SO_SNDTIMEO``` or not. If not, then do hand me a reproducer I can run and I'll poke some more at it. As mentioned, this might not be until next week however.

tchaikov:
@axboe sorry for the noise, and for the time you spent digging into it. You were right: it's not a kernel bug.

The failing sockets are accepted connections, which Seastar opens `O_NONBLOCK`. 

> Followup question - are you using `SO_SNDTIMEO`?

No SO_SNDTIMEO: the send carried MSG_DONTWAIT implicitly via the fd's O_NONBLOCK, not the SQE flags, so it only looked like "-EAGAIN without MSG_DONTWAIT".

My original wording was wrong. But while digging into it, I came across your commit https://github.com/torvalds/linux/commit/355afaeb578abac907217c256a844cfafb0337b2 states the principle for read/write

> if we hit -EAGAIN on a nonblock marked file ... it should just complete with -EAGAIN

, I noticed the behavior people keep hitting isn't spelled out in the liburing docs yet: 

> that io_uring honors the file's O_NONBLOCK and returns -EAGAIN instead of polling and retrying

so I've reworked the PR to document that, if you think it's worth having. Happy to close it otherwise.

PR #1605, opened 2026-06-23
test/reg-fd-only: Relax io_uring_submit() return value check

chrehrhardt:
For SQPOLL rings the man page of io_uring_submit(3) states that the return value can be higher than the number of submitted SQEs and on a very slow overloaded system we sometimes see the test fail with:

| short submit 16
| test 8 failed

To fix this, allow the return value to be higher than expected.


krisman:
Make sense to me.    We are submitting 32nops in batches of 8.  If sqpoll didn't see the full previous 8, __io_uring_flush_sq  can report a higher value.

PR #1620, opened 2026-07-21
clarify registered buffers iovec lifetime

jiv4:


axboe:
Commit message needs updating:

1) Signed-off-by (real identity, not some online handle)
2) Actual commit message. A commit message explains WHY a commit is being made.

See other git commits in the tree, they all should follow this rule. Few exceptions, but at least the title (first line) of the commit should be descriptive. And if that isn't enough, add the WHY in the body.

PR #1621, opened 2026-07-22
test/open-direct-link: fix sporadic -EBADF failure

Flowdalic:
The test runs consecutive test iterations that chain openat_direct, read, and close_direct operations. Some iterations, like test(1, 0, 0) where skip_success=1, configure the openat and close operations to skip producing CQEs upon success. As a result, the test suite only waits on the CQE generated by the read operation before proceeding to the next test run.

Because the previous iteration's close_direct might still be in-flight or deferred to task_work, and all iterations were hardcoding file index 0, a race condition occurs: an asynchronous close_direct from a prior test run can inadvertently close the newly installed file from a subsequent test run. This causes the newly submitted read operation to unexpectedly fail with -EBADF ("bad read -9").

Fix this by passing an incrementally unique file_idx to each test iteration. This isolates the test iterations and prevents them from racing over the same fixed file descriptor slot.

### DIFF
--- a/test/open-direct-link.c
+++ b/test/open-direct-link.c
@@ -16,7 +16,7 @@
 #define MAX_FILES	8
 #define FNAME		".link.direct"
 
-static int test(struct io_uring *ring, int skip_success, int drain, int async)
+static int test(struct io_uring *ring, int skip_success, int drain, int async, int file_idx)
 {
 	struct io_uring_cqe *cqe;
 	struct io_uring_sqe *sqe;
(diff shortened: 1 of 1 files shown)


Flowdalic:
Thanks for the quick merge. Could you please also have a look at https://github.com/axboe/liburing/pull/1617 and https://github.com/axboe/liburing/pull/1605 which fix also issues we are seeing in our CI setup. Thanks

axboe:
Sorry, a bit behind, vacation and still on vacation. Both those look good too, merged.

Flowdalic:
> Sorry, a bit behind, vacation and still on vacation. Both those look good too, merged.

No worries. Thanks for merging and enjoy your vacation.

PR #1624, opened 2026-07-28
queue: keep fallback timeout data alive until submission

### LINKED ISSUE #1164: Potential dangling pointer if io_uring_submit_and_wait_timeout() used with IORING_SETUP_SQPOLL
If using a kernel version prior to 5.11 and thus IORING_ENTER_EXT_ARG is not available, the `io_uring_submit_and_wait_timeout()` function falls back to submitting an OP_TIMEOUT operation to manage the timeout instead.

However, if you have enabled SQ-polling, then it's possible that this function might return without actually waiting for the OP_TIMEOUT SQE to have been processed by the kernel-thread if a completion-event is received before the timeout elapses.

As the OP_TIMEOUT SQE holds a pointer to the `__kernel_timespec` structure passed by the user to this function, if the function returns early and the caller has passed a pointer to `__kernel_timespec` value that was allocated on the stack, then it's possible that the `__kernel_timespec` memory will contain garbage by the time the kernel-thread processes the OP_TIMEOUT SQE and reads the timespec value.

I think in this case, the `io_uring_submit_and_wait_timeout()` function needs to wait until the kernel-thread has indicated it has consumed the OP_TIMEOUT SQE by advancing `head` beyond the SQ index. Otherwise, the caller cannot know when it is safe to free the `__kernel_timespec` structure that it passed to this function.

I think this could also potentially be an issue for non-SQPOLL uring instances that do not have the `IORING_SETUP_SUBMIT_ALL` flag set (which would be all pre-5.11 kernels, since this flag was only added in 5.18) in cases where the submission of an SQE fails. My understanding from the docs on this flag is that an earlier SQE submission failing will cause `io_uring_enter()` to return without submitting subsequent SQEs. And, as such a failure will post a CQE, it's possible that this could cause the `io_uring_submit_and_wait_timeout()` function to return before the enqueued `OP_TIMEOUT` operation is submitted, leaving it in the submission-queue with a pointer to a potentially-dangling `__kernel_timespec` object.

The `__io_uring_get_cqe()` function would need to continue calling `io_uring_enter()` to submit subsequent SQEs until the OP_TIMEOUT SQE has been processed.

soul-sol:
On kernels without IORING_FEAT_EXT_ARG, the timeout fallback places the caller-owned timespec pointer in an IORING_OP_TIMEOUT SQE. _io_uring_get_cqe() could return an already available CQE before SQPOLL consumed that SQE, or before a short non-SQPOLL submission reached it. The caller could then release stack storage while the ring still held the pointer.

Track the SQ tail containing the fallback timeout and keep submitting/waiting until the kernel SQ head reaches it. If a synchronous submission fails first, retract the last timeout SQE so a later submit cannot use the pointer. Other get-CQE paths retain their existing behavior and return values.

The regression test idles an SQPOLL thread, fills a 32K SQ without publishing it, forces the pre-EXT_ARG fallback, and verifies the call cannot return with unconsumed SQEs. It fails on current master with 32768 SQEs still ready and passes with this change.

Validation:
- Full liburing test suite on Linux 6.12: all tests passed
- ASan/UBSan: submit-and-wait, timeout, timeout-new, and defer-tw-timeout passed

Fixes #1164

### DIFF
--- a/src/queue.c
+++ b/src/queue.c
@@ -52,10 +52,12 @@ static inline bool cq_ring_needs_enter(struct io_uring *ring)
 
 struct get_data {
 	unsigned submit;
+	unsigned submit_tail;
 	unsigned wait_nr;
 	unsigned get_flags;
 	int sz;
 	int has_ts;
+	bool wait_for_submit;
 	void *arg;
 };
 
@@ -98,8 +100,16 @@ static int _io_uring_get_cqe(struct io_uring *ring,
 		}
 		if (sq_ring_needs_enter(ring, data->submit, &flags))
 			need_enter = true;
-		if (!need_enter)
-			break;
+		if (!need_enter) {
+			if (!data->wait_for_submit ||
+			    io_uring_load_sq_head(ring) == data->submit_tail)
+				break;
+			/*
+			 * SQPOLL submission is asynchronous.  Keep the caller's
+			 * submission data alive until the SQ thread consumed it.
+			 */
+			continue;
+		}
 		if (looped && data->has_ts) {
 			/*
 			 * When IORING_ENTER_EXT_ARG_REG is set, data->arg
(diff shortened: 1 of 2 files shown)


krisman:
Please read CONTRIBUTING.md.  Empty commit message, missing signed-off, commit must do one thing only, etc.

PR #1626, opened 2026-08-05
queue: add opt-in requeue helper for SQ_REWIND

carrerasdarren-cell:
`IORING_SETUP_SQ_REWIND` clears liburing's local SQ tail before entering the kernel. If a submission is short, the caller may either abandon the unconsumed entries or choose to retry them. Retrying requires relocating those entries to the beginning of the SQ.

This adds `io_uring_sq_requeue()` as an explicit opt-in helper. The application records `io_uring_sq_ready()` before submission, then passes that count and the submission return value to the helper before acquiring any new SQEs.

Submission-and-wait paths also stop internally resubmitting an unrelocated `SQ_REWIND` remainder. Without that guard, a short timed submission can submit already-consumed slots again.

The regression covers:

- the default abandon policy and explicit requeue policy;
- ordinary and fixed 128-byte SQEs;
- partial and fully failed submissions;
- timed submission with duplicate-completion detection;
- a pure wait with locally pending work; and
- the regular circular-ring behavior as a control.

Validation:

- Linux 7.0.0-28-generic
- normal GCC build and FFI symbol check
- ASan/UBSan build
- focused regression plus nine adjacent submission/SQ tests in both builds

### DIFF
--- a/man/io_uring_sq_requeue.3
+++ b/man/io_uring_sq_requeue.3
@@ -1,0 +1,41 @@
+." SPDX-License-Identifier: MIT
+.TH io_uring_sq_requeue 3 "August 5, 2026" "liburing-2.15" "liburing Manual"
+.SH NAME
+io_uring_sq_requeue \- requeue unconsumed SQEs after an SQ_REWIND submit
+.SH SYNOPSIS
+.nf
+.B #include <liburing.h>
+.PP
+.BI "void io_uring_sq_requeue(struct io_uring *" ring ","
+.BI "                         unsigned " submitted ", int " submit_ret ");"
+.fi
+.SH DESCRIPTION
+The
+.BR io_uring_sq_requeue (3)
+function relocates entries left by a short or failed submission on a ring
+created with
+.BR IORING_SETUP_SQ_REWIND .
+It moves the unconsumed entries to the beginning of the submission queue and
+restores the local queue tail so that a later submit can retry them.
+.PP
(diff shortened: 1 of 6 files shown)


krisman:
I don't think this should be done by default on a short submit.  We might not want to requeue, in which case relocating the requests would be a waste of time.  Perhaps the relocation could be exposed as a helper in io_uring and let userspace call it if they want.

carrerasdarren-cell:
Agreed. I updated the patch so a short `SQ_REWIND` submit is abandoned by default and added `io_uring_sq_requeue()` as an explicit opt-in helper.

I also stopped the submission-and-wait loop from internally submitting an unrelocated remainder, since that can resubmit the already-consumed prefix. The revised regression covers both policies, partial and failed submits, fixed 128-byte SQEs, and timed-submit duplicate detection. Normal and ASan/UBSan builds plus ten focused tests pass on Linux 7.0.

I force-pushed the revised single commit. I am happy to adjust the helper name or signature if you prefer a different API shape.

carrerasdarren-cell:
Following up now that the explicit opt-in requeue design has had some time to sit. The current single commit implements the requested userspace helper, avoids internal resubmission of an unrelocated SQ_REWIND remainder, and the focused normal plus ASan/UBSan test set remains clean. When convenient, could a maintainer take another look at the API shape and behavior? I am happy to make any project-preferred adjustment.

PR #1629, opened 2026-08-23
liburing: avoid signed overflow in buffer ring indexing

SantanDon:
## Summary

Avoid signed integer overflow in the public `io_uring_buf_ring_add()` helper when a high 16-bit ring tail is combined with a large positive `buf_offset`.

## Why

The helper computes the buffer slot with:

```c
(br->tail + buf_offset) & mask
```

`br->tail` is 16-bit and is promoted to signed `int` for the addition. With `br->tail = UINT16_MAX` and `buf_offset = INT_MAX`, UBSan reports signed integer overflow before the ring mask is applied.

The slot calculation is inherently modulo the ring size. Converting `buf_offset` to unsigned before the addition makes wraparound defined while preserving normal offsets and the existing masked index behavior.

## Verification

A focused UBSan harness using the current expression reports:

```text
runtime error: signed integer overflow: 65535 + 2147483647 cannot be represented in type int
```

With the proposed expression, UBSan is clean for offsets 0, 1, 2, 3 and the boundary case `UINT16_MAX + INT_MAX`; ordinary masked indices are unchanged. `git diff --check` is clean.

This is submitted as undefined-behavior hardening; I am not claiming an out-of-ring write or kernel exploit.


axboe:
Please follow the repo git format, you're missing a Signed-off-by.

PR #1634, opened 2026-09-03
[GIT PULL] Test: Introduce duration scale for upper timing bounds

zeehha:
Hey there,

as we run the tests on slower platforms like Qemu, we've introduced an environment variable `LIBURING_DURATION_SCALE` that specifies a factor to scale upper time bounds of tests where we ran into timing issues. A default factor of `1.0` is specified, so the timing of the tests stays as is.  

I'm sending here a first patch set that introduces the test library function and the scaling for a few tests. We have a few more of these patches that we could follow up with.

Best,
Chris

----
## git request-pull output:
```
The following changes since commit 4cf73437863c2e492d2a1d0f24330f391c0f075b:

  test/iowait.t: Skip if system is not quiesced for too long (2026-08-31 17:00:18 -0600)

  https://github.com/zeehha/liburing user/cge/duration-scale

for you to fetch changes up to fd5bf3fa9fdc39f5d877f823b3bb475ad972502f:

  test/mshot-shutdown-race: Scale period of hang detection (2026-09-03 16:03:20 +0200)

Chris Gellermann (4):
      test: Introduce duration scale for upper timing bounds
      test/min-timeout-wait: Scale upper time bound
      test/io-wq-exit: Relax thread exit deadline
      test/mshot-shutdown-race: Scale period of hang detection

 test/helpers.c             | 27 +++++++++++++++++++++++++++
 test/helpers.h             |  2 ++
 test/io-wq-exit.c          |  2 +-
 test/min-timeout-wait.c    |  2 ++
 test/mshot-shutdown-race.c |  4 ++--
 test/runtests.sh           |  1 +
```
----
<details>

## Pull Request Guidelines
```
```

2. Then an empty line.
```
```

certain specific format).

of truth, not github.

URL.

```
```

Use a real email address!

### Commit message example:
```

safe without synchronization.

```

</details>

----

### DIFF
--- a/test/helpers.c
+++ b/test/helpers.c
@@ -365,6 +365,33 @@ unsigned long long utime_since_now(struct timeval *tv)
 	return utime_since(tv, &end);
 }
 
+/*
+ * Scale a duration based on the environment variable LIBURING_DURATION_SCALE
+ * (default 1.0) that can be increased on slow platforms.
+ */
+unsigned long long t_scale_duration(unsigned long long duration)
+{
+	static float duration_factor = 0;
+	if (duration_factor > 0)
+		return (unsigned long long) (duration * duration_factor);
+
+	duration_factor = 1;
+	char *env_val = getenv("LIBURING_DURATION_SCALE");
+	if (env_val) {
+		char* endptr;
+		errno = 0;
+		float mul = strtof(env_val, &endptr);
+		if (mul <= 0 || errno)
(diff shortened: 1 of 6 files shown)


krisman:
I don't see a problem with this approach, but given that you need to touch the individual tests, I think it would also be fine and simpler to just increase the timeout of the failing tests, when confirmed the VM slowness it the real culprit for the failure.  Either way, looks good.

One thing, IIRC, runtests also has a global timeout for the tests.  does that fail as well? does it need to be scaled?

PR #1638, opened 2026-09-08
[GIT PULL] register: avoid overshooting the file descriptor limit

vpfkfl753:
File registration adds the requested table size to the current soft limit. With soft=64, hard=96 and a request for 80 files, the resulting limit of 144 is rejected and registration returns `-EMFILE`, even though 80 is allowed.

Set the soft limit to the requested size. The regression covers regular, tagged and sparse tables in separate child processes and checks that the hard limit is unchanged.

The new test fails in all three modes before the fix. Afterward, it and `file-register`, `file-update`, `file-verify`, `file-exit-unreg`, and `io_uring_register` pass on Linux 7.0.0-28-generic with GCC 13.3.

----
## git request-pull output:
```
The following changes since commit 4cf73437863c2e492d2a1d0f24330f391c0f075b:

  test/iowait.t: Skip if system is not quiesced for too long (2026-08-31 17:00:18 -0600)

  https://github.com/vpfkfl753/liburing.git fix-register-nofile-limit

for you to fetch changes up to 5730376586eb6bade05116a59f5e255c132bda04:

  register: avoid overshooting the file descriptor limit (2026-09-08 21:21:03 +0900)

Sungwon Woo (1):
      register: avoid overshooting the file descriptor limit

 src/register.c              |  2 +-
 test/Makefile               |  1 +
 test/file-register-nofile.c | 94 +++++++++++++++++++++++++++++++++++++++++++++
 3 files changed, 96 insertions(+), 1 deletion(-)
 create mode 100644 test/file-register-nofile.c
```
----
<details>

## Pull Request Guidelines
```
```

2. Then an empty line.
```
```

certain specific format).

of truth, not github.

URL.

```
```

Use a real email address!

### Commit message example:
```

safe without synchronization.

```

</details>

----


krisman:
I'm not sure this is a real issue at all.  Yeah, if you go above the hard limit it will fail.    You mention a few tests that fail, it is a matter of configuring something (probably systemd) to a sane default rlimit, no?  In which distro is this failing out-of-the-box, without any configuration?

Regarding the change, rlimit is global to the process.  If you limit the new rlimit to nr, a second ring or another file  openwould fail. It makes more sense to expand beyond the number of tags requested, which is exactly what was explicitly done by commit 98db8e88 ("src/register: improve RLIMIT_NOFILE increase").

PR #1639, opened 2026-09-15
configure: only run BPF clang check if clang is present

Flowdalic:
Commit f7f36b3beb15 added a check for clang, but still ran the
subsequent 'clang -target bpf ...' test unconditionally. If clang is not
installed, this causes configure to emit:

  ./configure: line 513: clang: command not found

Guard the test with '$has_clang = "yes"', and redirect the test
compilation output to config.log so failures don't leak to stderr.

Fixes: f7f36b3beb15 ("configure: check for clang as well for BPF")
Link: https://bugs.gentoo.org/982552
Signed-off-by: Florian Schmaus <flo@geekplace.eu>


krisman:
LGTM.

PR #1640, opened 2026-09-17
[GIT PULL] man: fix sq_thread_idle == 0 semantics in io_uring_sqpoll.7

SmileLi:
Documentation fix only; see the commit message for the kernel-side evidence.

----
## git request-pull output:
```
The following changes since commit 78dce99b660fd1caaf70ef6573a88bae9bbf6476:

  Merge branch 'harden/sanitize-second-pointers' of github.com:StressTestor/liburing (2026-09-11 13:00:18 -0600)

  https://github.com/SmileLi/liburing man-sqpoll-idle

for you to fetch changes up to 3f8a7a9bcd1cc6107317b228cf793d116ea8b761:

  man: fix sq_thread_idle == 0 semantics in io_uring_sqpoll.7 (2026-09-17 16:27:28 +0800)

ligj (1):
      man: fix sq_thread_idle == 0 semantics in io_uring_sqpoll.7

 man/io_uring_sqpoll.7 | 28 ++++++++++++++++++++++++----
 1 file changed, 24 insertions(+), 4 deletions(-)
```
----


man/io_uring_sqpoll.7
```diff
@@ -89,8 +89,16 @@ The
 .I sq_thread_idle
 field specifies how long (in milliseconds) the kernel thread will
 poll before going to sleep if no submissions are pending. A value of
-0 means the thread never sleeps (uses more CPU but provides lowest
-latency).
+0 is treated as unspecified, in which case the kernel uses a default
+of 1000 milliseconds. No value asks the thread to never sleep; if
```
krisman: What do you mean by "No value asks the thread to never sleep;" ?  How would you pass a no-value?  If you mean there is no support for the thread to never  but it just reads confusing. I suggest just deleting this sentence.

man/io_uring_sqpoll.7
```diff
@@ -89,8 +89,16 @@ The
...
+that is the desired behavior, pass a value large enough to cover the
+idle gaps of the workload, such as
```
krisman: I see what you mean by idle gaps, but is not a well-defined concept.  I suggest you just don't explain this workaround, because it is still not a guarantee. A correct application needs to watch for the IORING_SQ_NEED_WAKEUP and wake the thread if needed. Something like this would read better.

"To minimize latency, you can specify an arbitrarily large idle period at the expense of wasting cpu time by spinning the kernel thread waiting for new requests."

man/io_uring_sqpoll.7
```diff
@@ -213,8 +232,9 @@ appropriately for your workload.
 .IP \(bu
 .B Idle timeout tradeoff:
 A shorter idle timeout saves CPU but may increase latency when the
-thread needs to wake up. A longer timeout (or 0 for never sleeping)
-uses more CPU but provides consistent low latency.
+thread needs to wake up. A longer timeout uses more CPU but provides
+consistent low latency, and a timeout longer than the idle gaps of
+the workload keeps the thread from sleeping at all.
```
krisman: Not a guarantee.  It might be correct in practice, but an application shouldn't rely on it.  Say your application is suspended and then restarted.  The expected "idle gaps" will have elapsed and sqpoll halts.  A correct application needs to deal with this scenario.

krisman:
Hi,

Beyond the comments above, due to certificate of origin at least in the kernel, we require contributors to use their real name.  Not sure if Jens enforce this in liburing, though.

axboe:
We don't need real names, but a real identity, same as the kernel.

krisman:
> We don't need real names, but a real identity, same as the kernel.

Hm, there's a real distinction I didn't consider.   I didn't know it was explicit addressed by kernel comit d4563201f33a ("Documentation: simplify and clarify DCO contribution example language"), which is excellent.   Thanks for the correction.

